PR #11, opened 2015-07-22
Deref demand instead of comparing the pointer adr.

simonlynen:
De-reference the value of demand instead of comparing the pointer adr to zero. (If a pointer comparison to nullptr was intended, then this should use operator!=)


simonlynen:
Sorry, this is not the right fix

PR #95, opened 2015-09-29
Updated hash_map and hash_set to unordered_map and unordered_set for gcc.

pbomta:
The hash_map and hash_set are deprecated and need to be replaced.
Tested with gcc 4.8.4. and gcc 4.9.3.

### DIFF
--- a/src/base/hash.h
+++ b/src/base/hash.h
@@ -14,13 +14,10 @@
 #ifndef OR_TOOLS_BASE_HASH_H_
 #define OR_TOOLS_BASE_HASH_H_
 
-// Hash maps and hash sets are compiler dependant.
+// Hash maps and hash sets are compiler dependent.
 #if defined(__GNUC__) && !defined(STLPORT)
-#include <ext/hash_map>
-#include <ext/hash_set>
-namespace operations_research {
-using namespace __gnu_cxx;  // NOLINT
-}  // namespace operations_research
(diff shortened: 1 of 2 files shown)


lperron:
Tricky one.
Creating a dummy hash_xxx could be problematic for a client that import native hash_xxx and base/hash.h

I need to think about it.

pbomta:
Good point. I'd be happy to go through the library code and replace all usage of hash_xxx with unordered_xxx, as MSVC 2013 supports them (I can also test that). But I know nothing about STLPort. Is that even used nowadays?

lperron:
Unfortunately, it has to be done on our side.
But it is definitively on our radar.

Does it compile with a warning with gcc 4.9.x? Or does it refuses to
compile?

Thanks

00


2015-09-29 14:58 GMT+02:00 pbomta <notifications@github.com>:

pbomta:
It compiles with a warning.
But it's good to hear that it's on the todo-list.
Thanks.

daniel-j-h:
Reading the release notes for or-tools v6.0

> We have removed all usage of hash_map, hash_set in the C++ codebase as they are deprecated. They were replaced by unordered_map and unordered_set from the STL.

I guess this can be closed as resolved now.

pbomta:
Indeed! Thanks for the update.

PR #103, opened 2015-10-12
Allow users to pipe in the model via stdin

skosch:
Hi,

I've been running or-tools with FlatZinc files that are assembled by a script. In my situation, it's much more convenient to pass the code in via stdin instead of saving it to a file first. Here are the modifications I've been using on my end to make that possible. They may well not be up to snuff with your coding standards (sorry; I'm not a C++ guy), but I wanted to put them out there in case you guys are willing to add that feature in :)


lperron:
Thanks

PR #104, opened 2015-10-13
Update bitset.h

TheCoolRob:
The visual studio compiler throws an error when you apply a sign to an unsigned int. I think that this is equivalent functionality and will work for both VS and gcc.


TheCoolRob:
No problem. I've updated the branch - the changes should be reflecting now in this pull request.
Thanks

PR #112, opened 2015-11-04
[makefiles/PYTHON_INC] Allow non-standard location of python include files

totally:
Moves configuration of `PYTHON_INC` into Makefile.local, which allows overriding for python include files in non-standard locations via new `${UNIX,WIN}_PYTHON_INC` vars via Makefile.local (which is better than hacking Makefile.unix to get my use-case met).

I am sorry, I don't have a way to test on Windows.   `python.h` is found on centos disto I'm using, but the setup is non-standard to say the least.


totally:
Let me redo this with a better commit msg

Issue #63, opened 2015-08-05
Python 3 support


orwant:
Originally reported on Google Code with ID 47
```
I see on PyPI that your Python wrapper only works on Python 2. Any chance of adding
support for Python 3?
```

Reported by `kermit666` on 2014-08-21 08:35:40

glennon:
I feel like i should be moving faster toward 3 myself, so :+1: to keep the gravity moving that direction.

lperron:
Done.

Look for py3-or-tools on pypi.

Thanks

mjfwest:
I think you might want to look for py3-ortools and not py3-or-tools? 
https://pypi.python.org/pypi/py3-ortools/2.3408

lperron:
Corrected, thanks

lperron:
https://pypi.python.org/pypi?%3Aaction=pkg_edit&name=py3-ortools

hastebrot:
Python eggs for Windows are missing. Great work with Python 3 support!

lperron:
Yes,

I need to build them :-)

Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00


2015-12-14 12:47 GMT+01:00 Benjamin Gudehus <notifications@github.com>:

devster31:
I'm getting a strange error with `pip install py3-ortools`, pip can see it with `pip search` but when I install the package I get:
```
Collecting py3-ortools
  Could not find a version that satisfies the requirement py3-ortools (from versions: )
No matching distribution found for py3-ortools
```
I'm on OSX Yosemite, with `pyenv`, `python --version Python 3.5.0`.

lperron:
Sorry,

I built it for El Capitan.

setuptools cannot support two versions of OSX simultaneously.

--Laurent

Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00


2015-12-17 15:20 GMT+01:00 Daniele <notifications@github.com>:

devster31:
In this case I'm can build it from source, where can I find the instructions? Do I need to be careful of something in particular with pyenv or virtual environments?

lperron:
https://developers.google.com/optimization/installing

On mac,

install python3, checkout or-tools sources,

after make third_party

edit Makefile.local

add ADD_PYTHON_INC = -I/Library/Frameworks/Python.framework/Versions/3.5/
include/python3.5m/

Then run

  make install_python_modules
  make python
  make test_python

and then

[...]

PR #122, opened 2015-12-05
Get rid of appscript/gflags dependencies

cjdrake:
I see that @orwant filed #63 a while ago, regarding Python3 support. This commit is a baby step in that direction.

I recently asked @gpshead and @craigcitro about paths towards Python3 compatibility for or-tools. Since both google-appscript and gflags are Python2 only, and difficult to change, they suggested the path of least resistance might be to just get rid of them.

Please let me know what you think, and I can start making more changes like this.

### DIFF
--- a/examples/python/appointments.py
+++ b/examples/python/appointments.py
@@ -12,16 +12,22 @@
 # limitations under the License.
 #
 
-from google.apputils import app
-import gflags
+import argparse
+import sys
+
 from ortools.constraint_solver import pywrapcp
 from ortools.linear_solver import pywraplp
 
-FLAGS = gflags.FLAGS
-gflags.DEFINE_integer('load_min', 480, 'Minimum load in minutes')
-gflags.DEFINE_integer('load_max', 540, 'Maximum load in minutes')
-gflags.DEFINE_integer('commute_time', 30, 'Commute time in minutes')
-gflags.DEFINE_integer('num_workers', 98, 'Maximum number of workers.')
+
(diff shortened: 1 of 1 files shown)


lperron:
A few years ago, I ported or-tools to python3.
Using 2to3, everything went fine.

The blocking point was protobuf.

I need to check if protobuf 3.0.0 pre-release supports python3.

And we would like to keep gflags/google appcript in the code to simplify our python examples.

craigcitro:
protobuf3 supports python3, which is good going forward. i think it's unlikely that we'll get new gflags and apputils versions out any time soon, but your call.

cjdrake:
Cool, I'm glad we're all in agreement about porting to Python3.

The Python examples directory is an excellent resource for newcomers. Shouldn't we make it accessible to as many of them as possible, whether they are using Python 2 or 3?

Also, IMO, this kind of change actually makes this example *simpler*. Why? Because it gets rid of a dependency that very few people are familiar with, and replaces it with the de-facto standard argument parsing library from Python's standard library. Plus, it gets rid of whatever `app.run()` is implicitly doing behind the scenes, and replaces it with the more explicit system exit. As [pep20](https://www.python.org/dev/peps/pep-0020/) would say, "explicit is better than implicit.

lperron:
Hi, 

finally, I agree.
I have pushed a full cleanup that removed all dependencies on google-apputils and gflags.

Thanks for the suggestion.

You can delete this pull request.

cjdrake:
Done. Thanks :).

PR #150, opened 2016-01-29
Custom decision support in C#

supermihi:
Hi,

I've implemented a custom Decision in C# but ran into problems because ApplicationExceptions were being thrown. I seem to have solved the issue by copying the pattern found in NetDecisionMaker and NetDemon. Note that the attached code is not tested (compiling or-tools currently does not work on my machine).


src/com/google/ortools/constraintsolver/NetDecisionBuilder.cs
```diff
@@ -51,6 +51,55 @@ public virtual Decision Next(Solver solver)
...
+{
+  /**
+   * This method wraps the calls to Accept() and catches fail exceptions.
+   * It currently catches all application exceptions.
+   */
+  public override void AcceptWrapper(Solver solver)
+  {
+    try
+    {
+      return Accept(solver);
```
lperron: this is a void method. No return.

supermihi:
Hi Laurent,
thank's for spotting this. Now it should be correct ...

supermihi:
Sorry my first commit was rubbish ... mixed up accept and apply. ;-) Now, it compiles on my computer.

lperron:
Hi, 

I implemented and pushed the code myself.
You actually need to call ShouldFail() when catching exceptions.

I am not confident about memory management. That is I think you need to store the custom decisions in the decision builder otherwise they will be garbaged out.

supermihi:
Hi,

thanks for your answer. I already assumed that I have to keep custom decisions referenced in C# ... would it be possible to expose RevAlloc or is there too much SWIG magic going on that prevents this?

lperron:
I don't know. I guess not.
What is doable is to enrich NextWrapper() on the NetDecisionBuilder to
check the type of the decision, store it locally if it is a subclass of
NetDecision. Furthermore, this storage could be checked with a install for
RevInt() and cleaned if need be in the NextWrapper() too.

My 2 cents, swig is really too much magic for me. I just spent 2 weeks of
trials and errors to support protocol buffers in python/java/C# :-(

00


2016-02-05 9:07 GMT+01:00 Michael Helmling <notifications@github.com>:

supermihi:
Am 05.02.2016 09:16 schrieb lperron:
> I don't know. I guess not.
>  What is doable is to enrich NextWrapper() on the NetDecisionBuilder
> to
>  check the type of the decision, store it locally if it is a subclass
> of
>  NetDecision. Furthermore, this storage could be checked with a
> install for
>  RevInt() and cleaned if need be in the NextWrapper() too.
Ok ... if I'll happen to need & implement this, be sure to receive 
another pull request. More likely though that I will switch my entire 
implementation to C++. ;-)
> 
>  My 2 cents, swig is really too much magic for me. I just spent 2
> weeks of
>  trials and errors to support protocol buffers in python/java/C# :-(
I agree. Still, I'm very happy that you finally managed. Should make the 
bindings a whole lot more useful!

PR #200, opened 2016-06-29
Solow-Halim preprocessor to improve the number of iterations

Celebio:
This is an implementation of the Solow-Halim paper, Improving the Efficiency of the Simplex Algorithm Based on a Geometric Explanation of Phase 1 ([DOI 10.1504/IJOR.2009.025701](http://www.inderscienceonline.com/doi/abs/10.1504/IJOR.2009.025701))

I implemented it as a preprocessor.

As commented in the code, it modifies the problem by changing variables to begin (supposedly) geometrically near to the optimal solution.
For a maximization problem the changes are:
```
x_j' =  upper_bound(j) - x_j if cost(j) >  0
x_j' = -lower_bound(j) + x_j if cost(j) <  0
```

The effect of the preprocessor can be seen on netlib instances:
script : https://gist.github.com/Celebio/472f0e7a2e6f5505655a8df4889ba334
results : https://gist.github.com/Celebio/8a1e0cbaef29eb21c613154598310608

As there isn't always an improvement, the preprocessor is activated by an option, '-use_solowhalim". I think even small improvements can be hugely beneficial for usecases where lp solve is inside a loop.

Results also vary depending on the rank of `SolowHalimPreprocessor` call among other preprocessors. Placing it right after `DualizerPreprocessor` seems to be a good tradeoff : it doesn't give the best overall geometrical mean but the gain vary more depending on instances so the user may prefer to activate the option for a specific instance.

A few comments and potential sources of improvement:

- The condition `isFinite(ub) && isFinite(lb)` ([line 3465 preprocessor.cc](https://github.com/Celebio/or-tools/commit/0a1d9ed0c15b64b8c92b4347f6d20f169df89c49#diff-fafba723f9e84069d95a1df826192136R3465)) may be too strict. It should be possible to deal with the case where only one bound is finite.
- Obviously, instances with 'epigraphic' form (`max t, s.t. t ≤ cx, Ax=b`) can hardly take benefit of this. We could try to detect variables that intervene in the objective, but it may be computationaly expensive and unworthy.
- This implementation ignores MIP context. We could deal with integer variables by rounding variable offsets.

### DIFF
--- a/src/glop/parameters.proto
+++ b/src/glop/parameters.proto
@@ -351,4 +351,13 @@ message GlopParameters {
   // not create any OMP threads and will remain single-threaded.
   optional int32 num_omp_threads = 44 [default = 1];
 
+  // Whether or not use Solow-Halim preprocessor. It tries to begin the phase 1
+  // with a closer point to the optimal solution.
+  //
+  // Reference: Improving the Efficiency of the Simplex Algorithm Based on a
+  // Geometric Explanation of Phase 1
+  // https://weatherhead.case.edu/faculty/research/library/detail?id=12128593921
+  // DOI: 10.1504/IJOR.2009.025701
+  optional bool use_solowhalim = 46 [default = false];
+
 }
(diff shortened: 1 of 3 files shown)


Celebio:
I signed the CLA.

lperron:
Hi, 

this is very nice!

Thanks a lot for this contribution.

We believe this overlaps with one of our preprocessor, so in the future, we could extend one of our preprocessor to cover this one, and remove your code.
In the meantime, we will gladly keep it in or-tools.

Thanks again.

PR #205, opened 2016-07-05
fix for CMAKE_PLATFORM if no CMAKE_SUFFIX is set

KapecK:
This fix allowed build ortools in win x86.

Before thit fiix CMAKE_PLATFORM in x86 has one empty space:
"Visual Studio 12 2013 " 
and generates error:
"cmake  Could not create named generator Visual Studio 12 2013 "


KapecK:
I signed the CLA.

PR #264, opened 2016-10-29
Fixed bug with calling RoutingModel()

antoniomedrano:
Sending three int arguments gives a NotImplementedError: Wrong number or type of arguments for overloaded function 'new_RoutingModel'.
I put it to two arguments, not specifying a depot, which works.
Probably want to consider changing the C/C++ prototype to accept a depot argument.

### DIFF
--- a/examples/python/cvrptw.py
+++ b/examples/python/cvrptw.py
@@ -129,7 +129,7 @@ def main():
     # The number of nodes of the VRP is num_locations.
     # Nodes are indexed from 0 to num_locations - 1. By default the start of
     # a route is node 0.
-    routing = pywrapcp.RoutingModel(num_locations, num_vehicles, depot)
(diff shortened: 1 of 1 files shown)


furnon:
The new API always takes a depot argument in all languages. Are you
compiling from source or from a binary (the latter might be out of sync)?

dlahlou:
I'm afraid your python imports the ortools package from pypi instead of the one from the sources. It's a known issue for us, and we're looking for the best way to deal with it (PYTHONPATH doesn't solve this). For the moment, I recommend that you uninstall or delete the package installed from pypi.
P.S : you can use [inspect.getfile](https://docs.python.org/2/library/inspect.html#inspect.getfile) to check which module python imports.

antoniomedrano:
I installed via easy_install. So that version is different than building it on my machine?

PR #285, opened 2016-12-08
Update constraint_solver.swig

kennethalton:


kennethalton:
Fix problem where accessing null pointer in swig-gen addDimensionWithVehicleCapacity method.

PR #286, opened 2016-12-08
Fix problem with  accessing null pointer in swig-gen addDimensionWithVehicleCapacity method.

kennethalton:


kennethalton:
Fix problem with accessing null pointer in swig-gen addDimensionWithVehicleCapacity method.

PR #299, opened 2017-01-05
Maximization & Minimization functions collecting list input from third argument

aspencerpsu:
added another else statement if the expression is a list, to break apart variables and coefficients


aspencerpsu:
it's trivial, but for my gui and anyone else creating gui lp solvers, it's a nice feature to have variables and coefficients mapped for objective function. Please see my repository [for the pywraplp.py][1] file where the max and min functions are altered and the [simplex_class.py] file line's 60:69 for further clarification and explicit detail

jculpon:
This looks very stale, I'll take a look over the winter holidays; someone @google will need to officially assign this to me.

lperron:
Sorry, I do not get the point, or the added value over python iterations.

jculpon:
No worries then, I'm not sure there's "added" value in the sense of making this easier internally @google, so much as making the interface "friendlier";

PR #301, opened 2017-01-11
Fix whitespace chars in makefiles/Makefile.port

FloopCZ:
During the refactoring of makefiles, some TABs in Makefile.port have been mistakenly replaced by spaces, which made the project unbuildable. This commit fixes the errorneous whitespace chars in Makefile.port file.


FloopCZ:
I signed it!

PR #318, opened 2017-02-06
Bitcasts void* to integral type

daniel-j-h:
Makes GCC happy.

Alternative solution would be to cast to `intptr_t` and then to using the integral overloads.


daniel-j-h:
I signed it!

lperron:
Thanks

PR #319, opened 2017-02-15
Fix undefined behaviour found by -fsanitize=undefined with Clang.

PetterS:
None of these fixes seem too serious, but I had to do this to get a clean run without errors.


lperron:
Thanks

PR #340, opened 2017-03-11
Pull request separates nuget strategic steps

mwpowellhtx:
Steps including:

* Staging
* Packaging
* Distribution

Paused for the FlatZinc. Looks interesting, but will reserve that for a follow on effort, as it becomes necessary.

### DIFF
--- a/makefiles/Makefile.win.mk
+++ b/makefiles/Makefile.win.mk
@@ -21,6 +21,10 @@ STATIC_PRE_LIB = $(OR_ROOT)lib\\
 STATIC_POST_LIB = .lib
 O=obj
 E=.exe
+L=.lib
+DLL=.dll
+PDB=.pdb
+EXP=.exp
 OBJ_OUT = /Fo
 EXE_OUT = /Fe
 SWIG_LIB_SUFFIX = dll
@@ -44,11 +48,13 @@ ARCHIVE_EXT = .zip
 FZ_EXE = fzn-or-tools$E
 
 # Add some additional macros
+CD = cd
 ATTRIB = attrib
 TASKKILL = taskkill
 NUGET = nuget.exe
 NUGET_PACK = nuget.exe pack
 NUGET_PUSH = nuget.exe push
+NUGET_SRC = https://www.nuget.org/api/v2/package
 
 # Compilation macros.
 DEBUG=/O2 -DNDEBUG
--- a/tools/or-tools.nuspec
+++ b/tools/or-tools.nuspec
@@ -1,22 +1,22 @@
 <?xml version="1.0"?>
 <package >
-  <metadata>
-    <id>Google.OrTools</id>
-    <version>VVVV</version>
-    <authors>Google</authors>
-    <licenseUrl>https://github.com/google/or-tools/blob/master/LICENSE-2.0.txt</licenseUrl>
-    <projectUrl>https://developers.google.com/optimization</projectUrl>
-    <description>.NET Assembly for the OR-Tools project</description>
-    <copyright>Copyright 2016 Google</copyright>
-    <tags>OperationsResearch MathProgramming LinearProgramming ConstraintProgramming .NET</tags>
-    <dependencies>
-    <dependency id="Google.Protobuf" version="PROTOBUF_TAG" />
-    </dependencies>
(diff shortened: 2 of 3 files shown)


mwpowellhtx:
@lperron Please find the PR ready for merge. I resolved a conflict with a more specific ``CLR_ORTOOLS_DLL_NAME`` variable name, including ``CLR_PROTOBUF_DLL_NAME``, anticipating potential for ``CLR_FLATZINC_DLL_NAME``.

mwpowellhtx:
@lperron I am considering a patch for purposes of exposing solver serialization, i.e. between sessions, so to speak, but I don't want to get ahead of this PR for the time being. Thanks again for considering!

tools/or-tools.nuspec
```diff
@@ -1,22 +1,22 @@
...
-  </files>
+    <metadata>
+        <id>NNNN</id>
+        <version>VVVV</version>
+        <authors>Google</authors>
+        <licenseUrl>https://github.com/google/or-tools/blob/master/LICENSE-2.0.txt</licenseUrl>
+        <projectUrl>https://developers.google.com/optimization</projectUrl>
+        <description>.NET Assembly for the Operations Research Tools project</description>
+        <copyright>Copyright 2016 Google, Inc</copyright>
+        <tags>Operations Research Math Linear Constraint Programming C# .NET SWIG C++</tags>
```
lperron: AFAIK, tags are one word. Thus Math Programming will resolve to 2 tags without any link. 
I believe we are loosing something with this change. But I could be convinced otherwise.

makefiles/Makefile.win.mk
```diff
@@ -21,6 +21,10 @@ STATIC_PRE_LIB = $(OR_ROOT)lib\\
 STATIC_POST_LIB = .lib
 O=obj
 E=.exe
+L=.lib
```
lperron: Please define for unix too.

tools/or-tools.nuspec
mwpowellhtx: Could be my misunderstanding. I take *tags* to mean *keyword*, but I could be wrong about that. So if someone puts *Math* or *Constraint* in their tags, then the result would include ``Google.OrTools``. Not unlike HTML keywords meta tags for search engine optimization purposes.

makefiles/Makefile.win.mk
mwpowellhtx: **nix* is not in my purview right at the moment. I am focused on Windows and Windows only right now. Would a separate commit work?

tools/or-tools.nuspec
mwpowellhtx: From the [nuspec File Reference](http://docs.microsoft.com/en-us/nuget/schema/nuspec#optional-metadata-elements), *A space-delimited list of tags and keywords that describe the package and aid discoverability of packages through search and filtering mechanisms.* They really are not intended to capture further semantic meaning than that. That's what your wiki or other product reference material is there for.

mwpowellhtx:
@lperron If nothing further substantive, then please accept the merge. Thanks! Good stuff to follow.

makefiles/Makefile.win.mk
mwpowellhtx: @lperron One thought occurred to me, is the *csharp* invoked by the *unix* make files? If so, no problem.

tools/or-tools.nuspec
lperron: OK

makefiles/Makefile.win.mk
lperron: yes. I usually build and test csharp on unix first.
Thus my reservations about previous renaming.

lperron:
I fixed the unix build. Thanks

makefiles/Makefile.win.mk
mwpowellhtx: @lperron Ah, I see. IMO, most folks I work with understand *.lib* to mean static library, which ``L`` is a decent shorthand consistent with ``E``. Which, in Windows anyway, ``DLL`` is always *dynamic link*; except for **nix*, which ``.so`` is for *shared object library*.

That said, thanks for merging! :+1:

PR #344, opened 2017-03-16
Memory leak fix KnapsackSolver; for java proxy

Nielsdv:
Because the destructor is not disclosed the memory does not get freed. Added to configuration file and tested thoroughly.


Nielsdv:
I signed it!

PR #347, opened 2017-03-22
Use std::isnan

drhops:
Fixes following build errors on Ubuntu 16.10:
src/lp_data/lp_data.cc:39:28: error: ‘isnan’ was not declared in this scope


lperron:
Thanks

PR #348, opened 2017-03-24
Provided more elaborate Solver exportation capabilities

mwpowellhtx:
Also re-factored exportation apart from the process constraints as a help with increased code reuse.

@lperron Thanks for bearing with me exploring whether to expose ConstraintSolverParameters. Possibly there is a path there to obsolete export_file(), but I will leave that for future consideration.

For now, please find this PR with changes that support caller driven exportation.

### DIFF
--- a/src/constraint_solver/constraint_solver.cc
+++ b/src/constraint_solver/constraint_solver.cc
@@ -1665,20 +1665,10 @@ void Solver::ProcessConstraints() {
     ModelVisitor* const visitor = MakeStatisticsModelVisitor();
     Accept(visitor);
   }
-  const std::string export_file = parameters_.export_file();
-  if (!export_file.empty()) {
-    File* file;
-    if (!file::Open(export_file, "wb", &file, file::Defaults()).ok()) {
-      LOG(WARNING) << "Cannot open " << export_file;
-    } else {
-      CpModel export_proto;
-      ExportModel(&export_proto);
-      VLOG(1) << export_proto.DebugString();
-      RecordWriter writer(file);
-      writer.WriteProtocolMessage(export_proto);
-      writer.Close();
-    }
-  }
+
+  // TODO: TBD: much of the ExportModel has been re-factored; potentially could re-think
+  // the whole ExportFile portion of the ConstraintSolverParameters altogether.
+  ExportModel();
 
   if (parameters_.disable_solve()) {
     LOG(INFO) << "Forcing early failure";
--- a/src/constraint_solver/constraint_solver.h
+++ b/src/constraint_solver/constraint_solver.h
@@ -980,17 +980,54 @@ class Solver {
   // Abandon the current branch in the search tree. A backtrack will follow.
   void Fail();
 
-  // Exports the model to protobuf. This code will be called
-  // from inside the solver during the start of the search.
-  void ExportModel(CpModel* const proto) const;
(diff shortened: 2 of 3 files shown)


mwpowellhtx:
Addresses [ExportModel elaboration](http://github.com/google/or-tools/issues/349) issue.

lperron:
A bit too  intrusive to my taste :-(

I have changed a bit the export API to have the CP model as a return
parameter.
I also swiged it in C#

This way you can write:

    Solver solver = new Solver("TestConstraint");
    IntVar x = solver.MakeIntVar(0, 10, "x");
    IntVar y = solver.MakeIntVar(0, 10, "y");
    solver.Add(x + y == 5);
    CpModel model = solver.ExportModel();

And serialize the model any way you like using the C# protobuf API.

I hope this fits the bill.

00


2017-03-24 11:30 GMT-04:00 Michael Powell <notifications@github.com>:

mwpowellhtx:
@lperron Well,  I'd like to see it in action before signing off on it. Is the change committed yet that I can pull and check it out? I insist on at least being able to specify the file name apart from the parameters. I've got a similar thing in the works for the LoadModel. Thanks!

mwpowellhtx:
@lperron Notwithstanding passing file name in, an obvious next question is this: this yields the CpModel protobuf only? How then to actually save it to file? The problem is with SWIG, we lose visibility of the underlying C++-iness, class relationships, etc.

I will close this PR, but please expect another one addressing these concerns.

Thanks!

lperron:
No, the protobuf you get is in C#

See: https://developers.google.com/protocol-buffers/docs/csharptutorial

especially the parsing and serialization section.

I hope this helps

--Laurent

00


2017-03-24 14:50 GMT-04:00 Michael Powell <notifications@github.com>:

mwpowellhtx:
@lperron No it doesn't, because the type that is seen masking the CpModel has no knowledge anything in the Google.Protobuf library. That's all under the hood in the core C++ API.

mwpowellhtx:
@lperron Side note, I don't mind exposing this via the SWIG-ified C++ API as long as we can get to it. This is the limitation of SWIG, as cool as it is, we'll work within those constraints.

lperron:
Actually, I do generate the c# protobuf code in
src/gen/com/google/ortools/constraint_solver/Model.g.cs
It may not be a pure .NET code, (which I doubt), but it is not linked in
any way with swig.

And the serialization API do work in C#, as stated in the above
documentation.



00


2017-03-24 15:00 GMT-04:00 Michael Powell <notifications@github.com>:

mwpowellhtx:
@lperron I am building just now. I will circle around either with praise, or critique, as the case may be. Giving you the benefit of the doubt here, but I've got questions.

mwpowellhtx:
@lperron Okay, so a little feedback. It works great! Thank you. I believe it is even saving the name for variables, constraints, etc. That's super important especially when re-loading and aligning with model expectations.

mwpowellhtx:
@lperron Thinking out loud, I suppose one must have at least one visitor to capture a solution, correct? Then possibly also the decision builder? In other words, there would never be a case where we only had a decision builder?

Okay, this will satisfy the *WriteTo* use case. Now on-wards to a *ReadFrom* (?) use case, provided there is such a thing.

PR #351, opened 2017-03-25
Python Example was using constraints to create variables

darkstego:
Cleaned up the python example that was using constraints to create variables and adding redundant variables to decision builder. The new code search tree and speed improved by more than a factor of 10.


darkstego:
I signed it!

Issue #358, opened 2017-03-31
[Constraint Solver] CpModel proto exported after started search re-loads incorrectly


mwpowellhtx:
I see the [commit](http://github.com/google/or-tools/commit/14b18ba11aab87422d8694d5da9619ad27977020) testing what I think is an oversimplified use of an exported proto model.

However, in production, in my use case in particular, export would occur during the search, or perhaps after breaking the search (i.e. during NextSolution, after the NewSearch/EndSearch boundaries).

Following which, just prior to running a new search, would re-load an existing proto if it existed. But which a search may been in the process of already running.

Furthermore, in addition to model, I think the proto should include stateful restoration; i.e. we should not be revisiting old ground, previously determined solutions.

Anything short of that is simply academic in my view and is not useful.

mwpowellhtx:
The case being stated, I have also noticed that the re-loaded model breaks, is not restored in the same condition as when it was first arranged, after a search has begun running.

I will make PR that illustrates this fact.

mwpowellhtx:
Moreover, I think that's a bug rooted in how IntVar's (or derivatives) are being treated generically as IntExpr. Although technically IntVar derives from IntExpr, the IntVar-iness should be preserved. Or BooleanVar, and so on.

mwpowellhtx:
To be clear, the current example is academic in the sense that if all we're doing is exporting after arranging the model, constraints, etc, then running, this *doesn't add anything* to the equation. Why wouldn't we just pick up the *already-arranged* solver and run with *that*? Proto *doesn't add any value* to that picture. However, *correct restoration of a serialized solver state*, *does*.

mwpowellhtx:
Pulled latest code from the repo. From what I can determine, definitely seems to be an improvement. However, it is still a bit inconsistent for my taste.

Here is output from *NUnit* testing. I will also be updating the *testcp.cs* accordingly with some more elaborated ``Check``(s).

```
  Expected string length 1010 but was 866. Strings differ at index 152.
  Expected: "..."Equal", "value", "expression", "variables" ], "expression..."
  But was:  "..."Equal", "value", "expression" ], "expressions": [ { "name..."
  --------------------------------------------^

   at NUnit.Framework.Assert.That(Object actual, IResolveConstraint expression, String message, Object[] args)
   at NUnit.Framework.Assert.That(Object actual, IResolveConstraint expression)
[...]

PR #359, opened 2017-04-02
Commits corresponding with exportation expectations

mwpowellhtx:
Did a [little work](http://github.com/google/or-tools/issues/358) corresponding to a couple of expectations I have around exportation.

### DIFF
--- a/src/constraint_solver/constraint_solver.h
+++ b/src/constraint_solver/constraint_solver.h
@@ -2913,13 +2913,10 @@ class Solver {
   // returns false if the solver is not in search at all.
   bool CurrentlyInSolve() const;
 
-  // Counts the number of constraints that have been added
-  // to the solver before the search.
-  int constraints() const { return constraints_list_.size(); }
-
-  //// TODO: TBD(MWP): may expose these collections eventually, but not at the moment. at minimum be able to reference constraints by name
-  //const std::vector<Constraint*>* get_constraint_list() { return &constraints_list_; }
-  //const std::vector<Constraint*>* get_additional_constraints_list() { return &additional_constraints_list_; }
+  // Counts the number of constraints that have been added to the solver before the search.
+  int constraint_count() const { return constraints_list_.size(); }
+  /// Counts the number of additional constraints that have been added during the search.
+  int additional_constraint_count() const { return additional_constraints_list_.size(); }
 
   // Accepts the given model visitor.
   void Accept(ModelVisitor* const visitor) const;
--- a/src/constraint_solver/constraint_solver.h
+++ b/src/constraint_solver/constraint_solver.h
@@ -252,6 +252,7 @@ struct DefaultPhaseParameters {
 //
 class Solver {
  public:
+  // TODO: TBD(MWP): obsolete: not to oversimplify the issue, but I think that there is NO VALID REASON, in my estimation, for IntegerCastInfo, when we simply treat the AST nodes as what they are. Let the polymorphism virtual tables pick up what they are and that should be that.
   // Holds semantic information stating that the 'expression' has been
   // cast into 'variable' using the Var() method, and that
   // 'maintainer' is responsible for maintaining the equality between
@@ -2915,6 +2916,10 @@ class Solver {
   // Counts the number of constraints that have been added
   // to the solver before the search.
   int constraints() const { return constraints_list_.size(); }
+
+  //// TODO: TBD(MWP): may expose these collections eventually, but not at the moment. at minimum be able to reference constraints by name
+  //const std::vector<Constraint*>* get_constraint_list() { return &constraints_list_; }
+  //const std::vector<Constraint*>* get_additional_constraints_list() { return &additional_constraints_list_; }
 
   // Accepts the given model visitor.
   void Accept(ModelVisitor* const visitor) const;
--- a/src/constraint_solver/io.cc
+++ b/src/constraint_solver/io.cc
@@ -62,6 +62,8 @@ class FirstPassVisitor : public ModelVisitor {
   void BeginVisitModel(const std::string& solver_name) override {
     // Reset statistics.
     expression_map_.clear();
+    //// TODO: TBD: I'm not sure that expressions/variables, aren't being confused during AST visitation
+    //variable_map_.clear();
     delegate_map_.clear();
     expression_list_.clear();
     constraint_list_.clear();
@@ -81,6 +83,9 @@ class FirstPassVisitor : public ModelVisitor {
 
   void VisitIntegerVariable(const IntVar* const variable,
                             IntExpr* const delegate) override {
+    variable->Accept(this);
+    //// TODO: TBD: I think this is at least potentially a confusion of visited terms...
+    //delegate_map_[variable] = variable;
     if (delegate != nullptr) {
       delegate->Accept(this);
       delegate_map_[variable] = delegate;
@@ -154,6 +159,9 @@ class FirstPassVisitor : public ModelVisitor {
   }
 
   // Export
+  //const hash_map<const IntVar*, int>& variable_map() const {
+  //  return variable_map_;
+  //}
   const hash_map<const IntExpr*, int>& expression_map() const {
     return expression_map_;
   }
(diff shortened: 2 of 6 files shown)


mwpowellhtx:
@lperron I added a couple of interfaces to the API and elaborated on the testing a bit as well. Hopefully that helps a bit.

I'm somewhat on the fence as to whether a fully stateful load is even practical considering how the decision builder navigates the solutions, i.e. a tree (?). I imagine that once a branch has been visited, and subsequently rolled back, that any history of that visitation, good, bad, or indifferent, is lost.

I'm just a bit dubious on the front of whether exporting models at all would be that useful otherwise. Or perhaps I just haven't met that use case yet.

Anyhow, hope this PR finds you well.

src/constraint_solver/constraint_solver.h
```diff
@@ -252,6 +252,7 @@ struct DefaultPhaseParameters {
 //
 class Solver {
  public:
+  // TODO: TBD(MWP): obsolete: not to oversimplify the issue, but I think that there is NO VALID REASON, in my estimation, for IntegerCastInfo, when we simply treat the AST nodes as what they are. Let the polymorphism virtual tables pick up what they are and that should be that.
```
lperron: you are missing the point. But this issue is too complex, and is not a question of AST, nor virtual table.

src/constraint_solver/io.cc
```diff
@@ -180,6 +188,16 @@ class FirstPassVisitor : public ModelVisitor {
   }
 
  private:
+  //// TODO: TBD: not sure we are confusing terms with IntExpr, losing IntVar-iness, nevermind Boolean, etc...
```
lperron: We are not loosing intvar-ness, not boolean. This is taken care in the Solver::IntVar() method

src/constraint_solver/io.cc
```diff
@@ -2291,13 +2310,28 @@ bool CpModelLoader::BuildFromProto(const CpSequenceVariable& proto) {
...
+}
+
 IntExpr* CpModelLoader::IntegerExpression(int index) const {
   CHECK_GE(index, 0);
   CHECK_LT(index, expressions_.size());
   CHECK(expressions_[index] != nullptr);
   return expressions_[index];
 }
 
+IntExpr* CpModelLoader::IntegerExpressionByName(const std::string& name) const {
```
lperron: I do not want to store names in a hash map. This can be done at the client level.

lperron: or worse have a linear scan, so definitely in the client.

src/constraint_solver/constraint_solver.h
```diff
@@ -2916,6 +2917,10 @@ class Solver {
   // to the solver before the search.
   int constraints() const { return constraints_list_.size(); }
 
+  //// TODO: TBD(MWP): may expose these collections eventually, but not at the moment. at minimum be able to reference constraints by name
+  //const std::vector<Constraint*>* get_constraint_list() { return &constraints_list_; }
```
lperron: I definitely do not want to expose this., Especially the additional constraint list.

lperron:
As I said in previous discussion, I do not believe you can export/restore model without external knowledge, and I am on the fence of killing this feature altogether.

I believe that any practical application that need serialization will be better off storing data, and a function to reconstruct the model from the data. 

I am pretty sure I will not implement exact checkpointing of the search state, as this will freeze the solver, and I believe it can be better done using techniques from the embarrassingly parallel search.

src/constraint_solver/io.cc
mwpowellhtx: @lperron Name is already being stored. That's what ``SetName`` and ``Name`` buy us. If a client doesn't use this API, great. But when we do want to use it, it will be there. Should be transparent to either interest.

src/constraint_solver/io.cc
mwpowellhtx: @lperron Laurent, I ran a unit test that [proves otherwise](http://github.com/google/or-tools/issues/358). Furthermore, the fact that ``IntVar``'s are being *delegated* to ``IntExpr``, underscores this fact. So... What would it take ***not*** to delegate to ``IntExpr``?

src/constraint_solver/constraint_solver.h
mwpowellhtx: @lperron I can understand that. I'll push a commit that reverses this particular request. Do you have a problem with a by-name interface, at any rate?

mwpowellhtx:
@lperron FWIW, I'm starting to reconsider whether it is practical to do something like full restoration of state. However, approaching that question, I've found these interfaces useful. Will follow up with another commit. Thanks!

src/constraint_solver/io.cc
lperron: Can you copy paste which test shows that?

Not delegating is years of work, and lots of slowdown in the solver :-) So it won't change :-)

mwpowellhtx: @lperron I added a test that shows that. Merge the PR please. Thank you.

mwpowellhtx: @lperron As for the delegation part, the only place the delegation function is being used is then. When ``IntVar`` is delegated as an ``IntExpr``. Not to discount the contributions that went into that, but maybe that's a couple of hours thinking it through? Maybe a few more following that through; i.e. rename the function or comment it out, and see what breaks. That would provide some indication as to the scope. Notwithstanding why do we need to delegate an ``IntVar:IntExpr`` to ``IntExpr`` at all?

lperron:
Actually, on the test, if you remove constraint naming, and name the 2 solvers with the same name, they run OK.

src/constraint_solver/io.cc
lperron: I do not see your point

IntVar is a subclass of IntExpr.
The CP Solver creates IntExpr to represent arithmetic expressions (mostly)

expr = var_x + var_y

And expression is a min/max objects and offers events when the min or the max changes.
Constraints can subscribe to these events, and in many cases, this is enough.

IntVar are a subclass of IntExpr, it offers the same interface on min/max + value, removeValue and a few more methods.
BooleanVar are a special case of IntExpr and it should be invisible to the user.

Some constraints use the extended API (IntVar) and require thus variable as input.
The question is how to use x+y in a constraint (or any IntExpr).

You can create a special constraint Delegate(IntVar. IntExpr) that performs the z == x + y cast.
Then you can use z in the above constraint.

My point is that there are special cases delegate that are way more efficient that creating the expression first.  This is performed by the Var() call, and is registered in the CastInfo structure.

In the loader, it just tells you that the variable was created from the expression.
I do not understand how it affects you. 

You export the model, store the search information somewhere by name, load the model, scan to find the variables and reconstruct the search monitors.

And so far, in the 2 test I ran (SimpleTest and SimpleTestAfterSearch), expect the fact that I do not restore constraint names, they are working fine.

mwpowellhtx:
@lperron That's not what the error report indicates.

```
  Expected string length 1010 but was 866. Strings differ at index 152.
  Expected: "..."Equal", "value", "expression", "variables" ], "expression..."
  But was:  "..."Equal", "value", "expression" ], "expressions": [ { "name..."
  --------------------------------------------^
```

The difference begins with the original model including *variables*, whereas the loaded model does not.

mwpowellhtx:
@lperron Alright, it's making a little more sense to me now. It seems that whatever state is configured by ``NewSearch`` and/or during ``NextSolution``, is resolved and/or cleaned up afterward via ``EndSearch``. After that the exported model state seems good. Notwithstanding visited solutions, which we've discussed. :+1:

mwpowellhtx:
@lperron I pushed a patch to the unit test with the clarifications in mind.

All in all, I'm not sure this discovery process answered my original questions, per se, but it has helped clarify something about where things are currently.

I'm satisfied with that, for now. On wards...

mwpowellhtx:
@lperron I'd like to get this one resolved and move forward. What's the block for not merging?

mwpowellhtx:
@lperron Closing this one and will resubmit with a more focused PR. Thanks!

PR #367, opened 2017-04-17
Implement RecordIO in Java.

ThatRfernand:
- Use the same magic number for portability.
- Idiomatic method signature (void and exceptions instead of boolean)
- Provided as a Maven artifact


lperron:
Not accepting this patch. Now, you can write directly proto to file. No need for a record IO that was never truly supported.

PR #405, opened 2017-06-06
Ignore redundant dependencies, specify target for protoc binary

ambasta:
Hi @lperron 
Can you try this PR and see if it fixes the dependency issue on OSX?


ambasta:
Hi @lperron 

Can you try this PR locally before merging and let me know if it resolves the build issues you were facing?

PR #427, opened 2017-07-03
CMake Improvements

ambasta:
What's new:

- Added the ability to force rebuild of dependencies for C++ project
- Python project now correctly depends on the C++ project when C++ project is being built
- Project export cmake config files for detection in other dependencies


ambasta:
Btw, the python project can now be compiled and installed entirely using pip when published using the setup.py generated by CMake. i.e.
`mkdir build && cd build && cmake .. -DBUILD_PY=ON && make`
This would generate setup.py which can be used to create a release tarball as
`python setup.py sdist`
Finally once published,
`pip install tarball` 
(or ortools if published to pypi) would build the required swig modules automagically :)

PR #462, opened 2017-08-31
Docker image to build manylinux1 wheels

nazavode:
I'm using or-tools for a project, and the lack of linux wheels on PyPI makes the whole development and deployment process a real pain when you have to deal with multiple distros (and _obviously_ the production machine sits on the most ancient distro you can think of). This fork adds a docker image based on the [manylinux](https://github.com/pypa/manylinux) one (`quay.io/pypa/manylinux1_x86_64:latest`) and a [script](https://github.com/nazavode/or-tools/blob/master/tools/docker/build-manylinux1.sh) that, for each platform supported by `manylinux1`, builds, tests and exports binary wheels that are compliant with [PEP 513](https://www.python.org/dev/peps/pep-0513/). This has been tested with the current head (`6.3.4446`), the only thing that makes the whole process not reentrant is that the `quay.io/pypa/manylinux1_x86_64` docker image provides only the `latest` tag so any incompatible change that happens in there would break everything here, but for the sake of building portable wheels by hand looks fine to me.

### Usage

I've added a proper [target](https://github.com/nazavode/or-tools/blob/0f2ce6fb01376cdf984e161c13207fb3271f677e/tools/docker/Makefile#L159) to the `Makefile`, so in order to build and export all the `manylinux1` artifacts you need to:

```
$ cd tools/docker/
$ make manylinux1-pypi3
```

Out of the box, the target builds the following wheels:

```
$ ls -1 export/
ortools-6.3.4446-cp27-cp27m-manylinux1_x86_64.whl
ortools-6.3.4446-cp27-cp27mu-manylinux1_x86_64.whl
py3_ortools-6.3.4446-cp33-cp33m-manylinux1_x86_64.whl
py3_ortools-6.3.4446-cp34-cp34m-manylinux1_x86_64.whl
py3_ortools-6.3.4446-cp35-cp35m-manylinux1_x86_64.whl
py3_ortools-6.3.4446-cp36-cp36m-manylinux1_x86_64.whl
```

Python 2.6 (`cp26-cp26m` and `cp26-cp26mu`) is skipped by default due to some build problems (missing `Python.h` in some `Makefile` variable) but 2.6 is deprecated anyway, so I'm not looking into this. The list of platforms to be skipped is controlled by the [SKIP_PLATFORMS variable](https://github.com/nazavode/or-tools/blob/2e221ffc1d77da25e1dd914978d162e3a7f9bb1a/tools/docker/manylinux1.Dockerfile#L9).

### TODO

- [x]  I had to [patch](https://github.com/nazavode/or-tools/blob/0f2ce6fb01376cdf984e161c13207fb3271f677e/tools/docker/build-manylinux1.sh#L132-L145) the `makefiles/Makefile.python.mk` due to some errors and to disable the manual grafting of `libortools.so` and `RPATH` setting that are incompatible with the stuff that `auditwheel repair` does to a wheel. Probably it would be better to add a new target like `make pypi_archive` excluding the incompatible lines.
- [ ] Remove the `dummy_ortools_dependency.so` file from the wheels, I have to check but my initial guess is that it's not needed by portable wheels.

### DIFF
--- a/tools/docker/Makefile
+++ b/tools/docker/Makefile
@@ -42,7 +42,8 @@ pypi3: \
 	debian-9-pypi3 \
 	ubuntu-14.04-pypi3 \
 	ubuntu-16.04-pypi3 \
-	ubuntu-17.04-pypi3
+	ubuntu-17.04-pypi3 \
+	manylinux1-pypi3
 
 test: \
 	centos-7-test \
@@ -156,4 +157,4 @@ manylinux1-image:
 	docker build -f manylinux1.Dockerfile -t or-tools-manylinux1-image .
 
 manylinux1-pypi3: export manylinux1-image
-	docker run -w /root/or-tools -v `pwd`/export:/export or-tools-manylinux1-image:latest /bin/bash -c "/root/build-manylinux1.sh /root /export"
+	docker run -v `pwd`/export:/export or-tools-manylinux1-image:latest /bin/bash -c "/root/build-manylinux1.sh /root /export"
--- a/tools/docker/Makefile
+++ b/tools/docker/Makefile
@@ -157,4 +157,4 @@ manylinux1-image:
 	docker build -f manylinux1.Dockerfile -t or-tools-manylinux1-image .
 
 manylinux1-pypi3: export manylinux1-image
-	docker run -v `pwd`/export:/export or-tools-manylinux1-image:latest /bin/bash -c "/root/build-manylinux1.sh /root /export"
+	docker run -v `pwd`/export:/export or-tools-manylinux1-image:latest /bin/bash -c "/root/build/build-manylinux1.sh /root/src /root/build /export"
--- a/tools/docker/Makefile
+++ b/tools/docker/Makefile
@@ -8,7 +8,8 @@ images: \
 	ubuntu-16.04-image \
 	ubuntu-17.04-image \
 	centos-7-image \
-	debian-9-image
+	debian-9-image \
+	manylinux1-image
 
 images-no-cache: \
 	ubuntu-14.04-image-no-cache \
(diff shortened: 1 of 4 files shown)


nazavode:
@googlebot signed!

nazavode:
Changes to `makefiles/Makefile.python.mk` moved to a [patch file](https://github.com/nazavode/or-tools/blob/master/tools/docker/Makefile.python.mk.patch) that adds a new `pypi_archive_dir` target that stages all the files needed to build the wheel but, unlike `pypi_archive`, omits libraries grafing and `RPATH`s setting. Please note that the actual patching is now [in the Dockerfile](https://github.com/nazavode/or-tools/blob/2e221ffc1d77da25e1dd914978d162e3a7f9bb1a/tools/docker/manylinux1.Dockerfile#L63-L68).

lperron:
Thanks for doing this. I will need some time to audit your patch :-)

lperron:
Impressive

lperron:
Great, 

the next step is to remove the need for the patch.

I would clone the tools/setup.py to have a separate one without the dummy module which I also believe not necessary.  
Then build a separate makefile target .

Do you have an idea of the g++ version. Because an outdated version will cause problems too in the future

nazavode:
@lperron cheers. About the dummy dependency, I tried to remove it but `autidwheel repair` got pissed:
```
auditwheel.wheeltools.WheelToolsError: Cannot add platforms to pure wheel
```
That is because the `dummy_ortools_dependency.cc` was the only platform dependent bit of the distribution known to `setup.py` since it's added via `ext_modules` while the true platform dependent objects (all the `_pywrap*.so`) are added via `package_data`, so for setuptools those are generic files. I need to look into this, probably the best way would be to find the right `setup()` directive to add those wrappers to the package.

> Do you have an idea of the g++ version. Because an outdated version will cause problems too in the future

You are right, [a lot of people](https://github.com/pypa/manylinux/issues/118) is struggling with that ancient toolchain but unfortunately an update is not trivial. At the moment, there are two ways to cope with that:

1. graft into the wheels the newer `glibc` needed by the newer `C++` runtime to work properly **or**
2. wait for the `manylinux2` PEP :)

nazavode:
> the next step is to remove the need for the patch

...I can submit a PR for that with no hassle, it's up to you if you want to look into that.

nazavode:
@lperron submitted #468 that removes the patch file.

lperron:
Thanks. I will merge it.
I launched a interactive batch with the manylinux image. It reports gcc 4.8.2 which is good enough at the moment.

nazavode:
@lperron cheers!

Issue #473, opened 2017-09-06
C# - NodeEvaluator2 delegate can be GC'ed under memory pressure causing exception


DevNamedZed:
A callback was made on a garbage collected delegate of type `Google.OrTools!Google.OrTools.ConstraintSolver.NodeEvaluator2+SwigDelegateNodeEvaluator2_0::Invoke`. This may cause application crashes, corruption and data loss. When passing delegates to unmanaged code, they must be kept alive by the managed application until it is guaranteed that they will never be called.'

This issue exists in the current release (6.2.4388) and is not related to the NodeEvaluator2 array marshaling changes that were recently checked in.

This can be reproduced with the cscvrptw.cs. Adding a `console.writeline` to simulate memory usage to the distance evaluator causes repro each time on dev machine.

```cs
public override long Run(int first_index, int second_index)
{
    if (first_index >= locations_.Length ||
        second_index >= locations_.Length)
    {
        return 0;
    }

    Console.WriteLine(new string(' ', 500)); // Add here
    return (Math.Abs(locations_[first_index].x_ -
                        locations_[second_index].x_) +
            Math.Abs(locations_[first_index].y_ -
                        locations_[second_index].y_)) * coefficient_;
}
```

Will try to fix this tonight or tomorrow morning.

Thanks

lperron:
In ortools/com/google/ortools/constraintsolver/SolverHelper.cs, I am
already storing decision builder and search monitor when passing them on
the solver.
The same mechanism could be used for evaluators.

Thanks for doing this!!!

Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00


2017-09-06 4:09 GMT+02:00 DevNamedZed <notifications@github.com>:

DevNamedZed:
Hi Laurent,

It looks like SetVehicleCost was missing from the helper and also the new methods we exposed that take the NodeEvaluator[]. 

If its okay I changed the swig code to add the reference whenever it sees one of the types as a parameter. The public Aux methods will no longer be required, which is kinda good as the aux methods were not meant to be public called, but since the aux methods were public it is kinda of a break change. Please review and let me know if there any issues.

Thanks

lperron:
Thanks a lot for doing this.
I merged the pull request.

Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00


2017-09-06 10:23 GMT+02:00 DevNamedZed <notifications@github.com>:

DevNamedZed:
[...]

PR #475, opened 2017-09-07
fix DisownAndGetPermanentCallback not getting called on NodeEvaluator2 when array is used.

DevNamedZed:


lperron:
Thanks!!!

Mizux:
Related to #473

PR #559, opened 2018-01-19
Fsharp

acco32:


acco32:
I could if this is desired. My original intention was to leave the build of the library to the user. As this is dependent on the C# library, that makefile could be extended to include this as a build target. Perhaps rename C# to *dotnet* which would have child targets for each of those languages respectfully.

Mizux:
Ok, so I'll merge it "as it" !
Thanks again for the dotnet support !

Let's open an issue if we want to rework the dotnet languageS makefile integration.

PR #566, opened 2018-01-27
Integration into main makefile

acco32:
Add support to build F# library from main Makefile. Should use the existing keyfile for signing (since it depends on C# library).


lperron:
I have disabled the include in the Makefile.
Can you send me a merge request where nothing breaks if F# is not installed?

PR #573, opened 2018-02-08
.net standard changes

DevNamedZed:
create readme for .net standard
fix .net standard version and signing
update archive makefile to include .netstandard binaries and examples


lperron:
This is extremely nice, and a long standing TODO on my list.
Can you also document how to build or-tools from sources for dotnetcore?
And how to test it in that the same settings.

second question, do you know if we can upload multi-binary (mac, windows,linux) nuget packages to nuget.org

third, is it compatible with the current nuget package or should it be keep separated (Google.OrTools and Google.OrTools.Core)?

lperron:
can you add ortools/com/google/ortools/sat/*.cs to the list of included files.
I am adding the equivalent of the python cp_model.py layer here (CpModel.cs).

Thanks

DevNamedZed:
I added some documentation in regards to the build. I will try to add a test target for .net standard sometime soon.

It is definitely possible to do that. We need to combine the output from builds on multiple platforms though. Maybe can create a script that downloads the release zips from github and uses them to create the multi platform nuget. If using github to get the binaries is okay I will work on the script for this.

Let me know if there are any issues. Please give me a few days or so to get some time to add the test targets and write the script to create the multi-platform nuget.

Thanks

lperron:
Interesting. Combining binaries from github is OK. But low priority.
Currently I would like to know:
  - Can we can replace our windows .NET build with the .NET Core version and upload the new one to nuget?
  - Can we replace our mono builds with .NET Core on linux and mac?

Thanks again for the contribution

promontis:
Wow! Just wanted to say this is amazing!!!

promontis:
@lperron is there a nightly build so that we can start using this from nuget?

DevNamedZed:
Thanks for review and getting checked in.

There might be issues replacing the existing mono and C# builds. We are building a .net standard 2.0  assembly. It is only compatible with latest versions of mono and desktop .net framework. If its okay to drop support for older version then we can do this. I tested on mac, I will see if I can setup a linux computer to test with later.

https://github.com/dotnet/standard/blob/master/docs/versions.md

It would be awesome to have a single nuget that contained linux, osx, windows, etc binaries for x86, x64, etc. If there is a better place then github artifacts to get binaries we could do that instead. 

Thanks for creating this awesome library.

Thanks

lperron:
We have appveyor builds, but is consumes our cmake config, which does not
yet build C#, only python and C++.

It should be easy to build from sources.

00


2018-02-08 19:38 GMT+01:00 Michel van den Berg <notifications@github.com>:

DevNamedZed:
Awesome, I guess AppVoyer can make the multi platform nuget then

promontis:
From what it is worth, I would drop support for older versions; just replace it with .net standard 2.0.

Mizux:
I currently run appveyor from time to time to test the CMake ortools port
https://ci.appveyor.com/project/Mizux/or-tools
Hope to add a webhook and the badge on the README.md soon...

note: As @leperron said, I still need work to have the csharp/fsharp on CMake (On my current ToDo/InProgress)
note 2: we also have travis CI CMake jobs https://travis-ci.org/google/or-tools/branches
note3: I also have some cmake/Makefile docker based test to verify install work properly and can be use by an other cmake project...

promontis:
@DevNamedZed how did you make it work on the mac? I've tried running it using `make netstandard`, but it fails one generating the .snk file with the exception: 

```
Unhandled Exception: System.PlatformNotSupportedException: 'CspParameters' requires Windows Cryptographic API (CAPI), which is not available on this platform.
```

Should I run this on Windows and copy the files? Also, is it not yet compatible with .net core?

DevNamedZed:
Sorry about that. There is no sn.exe with .net core so I made a little command line program to generate. I am pretty sure I tested the key generation on a mac. I might have made a change on mac and not copied over to my maindev machine when I submitted the pr? Looks like if CLR_KEYFILE is not set then signing is skipped, can use as a workaround.

A platform specific .net standard 2.0 assembly is generated by the build. It should be compatible with every platform supported by both or-tools and .net standard 2.0. Per the discussion hopefully a multi platform nuget will be created too.

I am traveling right now and don't have a mac to test with. I will fix as soon as I get back in the next day or so.

Thanks

DevNamedZed:
I dont know how this was working before when I tested it but it seems like signing support for non-windows platforms was recently introduced.  Sorry about that.

https://github.com/dotnet/roslyn/issues/8210

I submitted a PR to disabled signing altogether for now. Maybe can turn back on with future version of dotnet sdk.

PR #586, opened 2018-02-16
dotnet standard changes

DevNamedZed:
disable signed build for .net standard
fix netstandard clean
fix dotnet sdk path detection
fix unsupported exception when generating key


Mizux:
Seems you are using `ziad@Ziads-Mac-mini.local` as email on your Mac so sign your patch
cf https://patch-diff.githubusercontent.com/raw/google/or-tools/pull/586.patch

flashfm:
Hi @DevNamedZed 
nice work!
have you ever managed running netstandard version on Linux?

lperron:
This is now obsolete as we have released dotnet core support in v6.8

Issue #569, opened 2018-02-05
solver returns an invalid solution for the vrptw with breaks


pmateusz:
A solution found by the constraint programming solver violates breaks of vehicles. The breaks are fixed and performed. A repro is included.

It is possible to define an instance of the Vehicle Routing Problem with Time Windows and Breaks for which the solver will report an invalid solution. Furthermore, the solver fails to recognize that the returned solution is invalid. A more appropriate action would be either dropping more visits or returning a nullptr solution.

Repro example is available in the fork: https://github.com/pmateusz/or-tools/commit/ff78b5c27d7db3a5b8c6ef1791ca3df35b55fd93

*Example output*
```shell
./cvrptw_with_breaks_repro
E0205 17:14:37.660873 11997 cvrptw_with_breaks_repro.cc:390] The time [13:00:00, 13:45:00] allocated for the visit (0 [12:15:00, 13:15:00] 00:45:00) overlaps with the break [13:00:00, 16:00:00] of vehicle (0)
E0205 17:14:37.661190 11997 cvrptw_with_breaks_repro.cc:390] The time [19:30:00, 19:45:00] allocated for the visit (8 [19:30:00, 20:30:00] 00:15:00) overlaps with the break [19:30:00, 20:00:00] of vehicle (17)
E0205 17:14:37.661250 11997 cvrptw_with_breaks_repro.cc:390] The time [19:00:00, 19:30:00] allocated for the visit (9 [19:00:00, 20:00:00] 00:30:00) overlaps with the break [19:00:00, 19:30:00] of vehicle (18)
```

Could you confirm the bug? Are you aware of any workarounds for the problem reported?

pmateusz:
Finally I didn't use the break feature from the routing library and implemented a custom constraint.

You may consider reviewing how the breaks are implemented currently. For example the logic for symmetry breaking does not work as expected, because the code is unreachable: https://github.com/google/or-tools/blob/d3bd654c287f8287cd83994f35bcff1d9144e595/ortools/constraint_solver/routing.cc#L4958

furnon:
Thanks for reporting. We will investingate the issue.

lperron:
Indeed. They were partially broken in 6.x.

PR #595, opened 2018-02-23
Example of a Vehicle Routing Problem with Uninterruptible Breaks

pmateusz:
An example of the Vehicle Routing Problem with Time Windows and Breaks. All breaks are mandatory and must be performed without interruptions. Since breaks without interruptions are not available off-the-shelf a custom constraint is developed. A warning message is printed every time vehicle's break overlaps with a visit.

Workaround for the issue: #569


pmateusz:
I have signed it!

Mizux:
Hi,

It seems "break" feature has been fixed internally, so I hope for v6.9 to bump routing library to our current internal code...

BTW I still keep this PR "open" until break is updated and this PR become obsolete.

pmateusz:
So it was a bug indeed. Thank you for confirmation! I am pleased that the work I did helped a little. It is of course always better to fix the root cause rather than rely on the workaround my PR provides. Feel free to close the PR once it is best.

tomika:
Dear Mateusz,
I would like to know if it is really neccessary to add
solver()->AddConstraint(solver()->MakeIntervalVarRelation(current_interval, Solver::STARTS_AFTER_END, last_interval));
constraints?
Is it not a basic, already existing constaint in the whole routing calculation?
Thanks for your answer in advance.

Issue #180, opened 2016-05-05
Multi-threading CBC solver


antoniomedrano:
Is there a way to multi-thread the CBC solver to use more than just one processor core? I'm using the python interface for or-tools.

hxd1011:
Hi, I have tried that with CBC stand alone executable. What I leaned is that, some builds may not have multi-thread enabled. i.e., -threads option is not always available.

Check this link and you will find all 2.7 builds are not parallel enabled (has -parallel in the name.)
http://www.coin-or.org/download/binary/Cbc/

Do not know How google OR will handle that

lperron:
When build CBC in make third_party, 
I am using this flags in the configure part

ADD_CXXFLAGS="-DCBC_THREAD_SAFE -DCBC_NO_INTERRUPT"


Should I add something else?

sschnug:
(Without knowing anything about Cbc's incorporation into ortools) I think in terms of configure ```--enable-cbc-parallel``` [might](https://github.com/h-g-s/lp/issues/1#issuecomment-357816478) be relevant.

rlloretb:
Just to make it clear, is the parallel option for CBC enabled by default?

lperron:
Definitely for unix

https://github.com/google/or-tools/blob/stable/makefiles/Makefile.third_party.unix.mk#L361

and I believe so for windows

https://github.com/google/or-tools/blob/stable/makefiles/Makefile.third_party.win.mk#L369
Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00

Issue #57, opened 2015-08-05
Cannot set (Linear)-SolverParameters or solver-specific parameters  for CBC in .Net


orwant:
Originally reported on Google Code with ID 41

```
I would like to solve a LP by using different parameters (especially different from
the default parameters). I tried to set some solver specific parameters for CBC by
two different ways:

1. I  tried to change the default values of the parameters LP_ALGORITHM, BARRIER and
DUAL by using the SetIntegerParam/SetDoubleParam-methods, but the values are unchanged
afterwards. The other default-parameters are changeable by using this methods.

2. Additionally, I tried to use SetSolverSpecificParametersAsString() with a string
of CBC-specific parameters. Currently, SetSolverSpecificParametersAsString() always
returns false, unless the input is empty. Could you give me an example to use this
method?

What version of the product are you using? On what operating system?
.Net OrTools.dll Version 3236 





```

Reported by `luisachristine` on 2014-04-10 09:10:26

maeehart:
This seems not to work on C++ either. Is anybody working on this?

The comments in the header linear_solver.h are not helpful since I cannot find any information about CBC configuration files. I have tried SetSolverSpecificParametersAsString("ratio 0.05") and SetSolverSpecificParametersAsString("-ratio 0.05"). Firs of these works from the command line and the second one from the interactive optimizer, but neither has any effect from Google or-tools.

Any help is appreciated. Also, there is a question about this at https://groups.google.com/forum/#!searchin/or-tools-discuss/cbc%7Csort:date/or-tools-discuss/pfjxg7bg0Rg/e2VJD82XCgAJ, with no answer, so I am not the only one struggling with this.

lperron:
I do not know how to do it in C++. Api is really unfriendly.

PR #637, opened 2018-03-30
Build CBC with --enable-cbc-parallel option

yingzong:
This enables multithreading support for 'unix' platforms and has
been tested to work on macOS.

However, project configs relating to Makefile.third_party.win.mk
have not been similarly modified.

Addresses issue #180.

### DIFF
--- a/makefiles/Makefile.third_party.unix.mk
+++ b/makefiles/Makefile.third_party.unix.mk
@@ -250,7 +250,7 @@ dependencies/install/bin/cbc: dependencies/sources/Cbc-$(CBC_TAG)/Makefile
 	cd dependencies/sources/Cbc-$(CBC_TAG) && $(SET_COMPILER) make -j 4 && $(SET_COMPILER) make install
 
 dependencies/sources/Cbc-$(CBC_TAG)/Makefile: dependencies/sources/Cbc-$(CBC_TAG)/Makefile.in
-	cd dependencies/sources/Cbc-$(CBC_TAG) && $(SET_COMPILER) ./configure --prefix=$(OR_ROOT_FULL)/dependencies/install --disable-bzlib --without-lapack --enable-static --with-pic ADD_CXXFLAGS="-w -DCBC_THREAD_SAFE -DCBC_NO_INTERRUPT $(MAC_VERSION)"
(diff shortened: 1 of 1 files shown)


yingzong:
I signed it!

Mizux:
Hi, thanks for the patch need to check how this option is managed by Cbc (i.e. do we need to enable pthreadish option too)

According to a quick (partial) grep
```sh
$ cd /dependencies/sources/Cbc-2.9.9
$ grep --color -rIin "\--enable-cbc-parallel" .
...
./Cbc/configure.ac:184:                [AC_MSG_WARN([--enable-cbc-parallel selected, but --with-pthreadsw32-lib not given])
./Cbc/configure.ac:191:                [AC_MSG_WARN([--enable-cbc-parallel selected, but --with-pthreadsw32-incdir not given])])
./Cbc/configure.ac:198:                 [AC_MSG_WARN([--enable-cbc-parallel selected, but -lrt unavailable; using gettimeofday instead of clock_gettime])])
./Cbc/configure.ac:203:                 [AC_MSG_ERROR([--enable-cbc-parallel selected, but -lpthreads unavailable])])
...
```

So maybe we should also check avaibility and provide pthread to Cbc

personal note: need to check if my [Cbc CMake port](https://github.com/Mizux/Cbc) is compiled with parallel stuff...

yingzong:
Almost certain `pthreadsw32` is win32 stuff, but same sort of check as `pthreads`, i.e. whether the system can create threads.

I understand the concern here -- the build shouldn't fail on account of this extra flag. In case the system can't create threads and you want the build to continue without that flag, I suppose you'd have to check the same in the `ortools` build itself and adjust the argument accordingly. It wouldn't be useful to pass `-lpthreads` since that assumes you can find the library and that it does exist. Alternatively, another way is to enable/read a key like `UNIX_CBC_PARALLEL` in `Makefile.local` and use that to decide whether to pass `--enable-cbc-parallel` in the 'unix' version.

pank3992:
Hi, I've built this version for python. But I can't see it using all the processes. Can you please clarify, if I need to make any changes in the code.

yingzong:
@pank3992 This change only enables multithreading support in the CBC library that ortools depends on. Additional parameters need to be passed at runtime to make use of it and ortools doesn't support that yet. I have an ongoing PR for that (#638), and there is an also a couple of tracking issues (#57, #672)

pank3992:
thank you @yingzong for the quick response.

Mizux:
Thanks to https://github.com/google/or-tools/commit/6daa710a8eaa80fa577247c079aafa18b1931909 it should be possible to enable multi thread.

Todo: write an example how to pass the parameter in all languages

PR #638, opened 2018-03-31
Add parameter passing to the CBC interface

yingzong:
Previously, CBCInterface does not read from solver-specific
parameters because there is no ReadParameterFile implemented.

This patch implements a ReadParameterFile that takes these
parameters, saves them in the CBCInterface instance, and passes
them to callCbc during Solve as additional arguments.

Tested to work with e.g. the "-threads" argument.

Addresses issue #57.

### DIFF
--- a/ortools/linear_solver/cbc_interface.cc
+++ b/ortools/linear_solver/cbc_interface.cc
@@ -370,7 +370,10 @@ MPSolver::ResultStatus CBCInterface::Solve(const MPSolverParameters& param) {
   // through callCbc.
   model.setAllowableFractionGap(relative_mip_gap_);
   // NOTE: Trailing space is required to avoid buffer overflow in cbc.
-  int return_status = callCbc(solver_specific_parameters_ + " -solve ", model);
+  std::string call_string = " -solve ";
+  call_string = solver_specific_parameters_ + call_string;
+  int return_status = callCbc(call_string, model);
+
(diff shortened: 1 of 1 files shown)


yingzong:
This failed in some Java test in `GLPK_MIXED_INTEGER_PROGRAMMING`. Likely not be related to this change. What's the process for continuing?

ortools/linear_solver/cbc_interface.cc
```diff
@@ -363,7 +370,7 @@ MPSolver::ResultStatus CBCInterface::Solve(const MPSolverParameters& param) {
   // through callCbc.
   model.setAllowableFractionGap(relative_mip_gap_);
   // NOTE: Trailing space is required to avoid buffer overflow in cbc.
-  int return_status = callCbc("-solve ", model);
+  int return_status = callCbc(solver_specific_parameters_ + " -solve ", model);
```
Mizux: not sure this space is safe

yingzong:
Fixed the problem. It wasn't the extra space(s), but (I believe) the move-construction performed by `std::operator+(string)` when there is a constant involved (see http://www.cplusplus.com/reference/string/string/operator+/). The moved string ` -solve ` was probably getting freed in multiple threads.

Mizux:
@yingzong I'll revert this commit see #672 for details

PR #655, opened 2018-04-18
FSharp updates

acco32:
Utilize the dotnet toolchain and be able to compile both netfx, netstandard, netcore from single project.  Current target is for netfx but the netstandard library can be imported into netfx project and be compiled (with the use of the netstandard import). Nuget package has both net462 and netstandard2.0 binaries.


acco32:
@Mizux dotnet core needs to be installed on build runner for latest changes. Where is that configuration?

ortools/fsharp/README.md
```diff
@@ -1,25 +1,12 @@
...
-
-let opts = SolverOpts.Default
-            .Name("Equality Constraints")
-            .Goal(Minimize)
-...
-```
+## Compiling a standalone binary
 
-One can also use the makefile found in the root folder to accomplish the same task.
+Use the makefile found in the root folder. IT to accomplish the same task.
```
Mizux: "IT" don't get it...

.travis/install.sh
```diff
@@ -42,7 +53,8 @@ if [ "${BUILDER}" == make ]; then
 				installmono
 			elif [ "${LANGUAGE}" == fsharp ]; then
 				installmono
-				sudo apt-get -yqq install fsharp
+        sudo apt-get -yqq install fsharp
+        installdotnetsdk
```
Mizux: indent use tab, I could fix it later if you want

ortools/fsharp/README.md
acco32: Don't get it either. Will remove.

.travis/install.sh
acco32: Fixed. What is going on with AppVeyor builds?

Mizux:
I should have fixed the compilation on master -> will it be possible for you to rebase your branch on master so we can validate with Appveyor CI that's everything is fine ?
If it's too complicated, let me know maybe we can blind merge it, it won't be the first time master is broken in case of failure...

acco32:
I rebased against master and pushed. Should be all good now.

Mizux:
Thanks !
The first two appveyors jobs (Makefile based) pass thanks again !

acco32:
Sure. Now that this uses the latest toolchain, you could probably unify all the .NET related languages. It would require some major changes to current directory structure. I saw there were a few other PRs related to dotnet. Not sure if anyone else has taken this on or if this is something you want to do on your end.

acco32:
gflags, glog, protobuf are already provided per the imported nuget package. With regards to the directory structure, it is similar to how I structured to FSharp folder. A solution file at the root project and projects as child folders. For unification, the top level solution file would remain and all the dotnet sources would be in child folders. 

I started this on a branch in my fork: dotnet. It compiles but still migrating the c# targets (which are scattered throughout the project). Will add a test project and the FSharp projects afterward.

Mizux:
Currently we have package configs in `<root>/tools/language` (e.g. csharp, netstandard) and `<root>/tools` (e.g. setup.py) except fsharp in `<root>/ortools/fsharp`.

I think lperron wanted to put them in `<root>/config/language`, or at least keep `<root>/ortools` only for source code (and swig wrapper)
For protobuf they use `<root>/language` (csharp, java, python, ruby).
Personally, I just want to remove the mess in `<root>/tools` and have consistent layout across languages. After nuget is a package like deb brew formula and csharp a language so still thinking about having for example `<root>/package/nuget` etc and a `<root>/fsharp` for example will also make sense...

PR #668, opened 2018-04-25
Add .Net Standard 2.0 support to OR-Tools

acco32:
@Mizux @lperron 

Able to build OrTools using [.NET Standard 2.0](https://docs.microsoft.com/en-us/dotnet/standard/net-standard). This makes it runnable in all frameworks (core 2.0, netfx 4.6.1) Includes C# & F# library. Still a work in progress but created PR to get some feedback. 

`make dotnet` will create the library and `make test_dotnet` will create and run the converted tests. Not all of them were converted.


makefiles/Makefile.dotnet.mk
```diff
@@ -0,0 +1,251 @@
...
+	"$(DOTNET_EXECUTABLE)" build "ortools$Sdotnet$S$(ORTOOLS_DLL_TEST)$S$(ORTOOLS_DLL_TEST).csproj"
+	"$(DOTNET_EXECUTABLE)" clean "ortools$Sdotnet$S$(FSHARP_ORTOOLS_DLL_TEST)$S$(FSHARP_ORTOOLS_DLL_TEST).fsproj"
+	"$(DOTNET_EXECUTABLE)" build "ortools$Sdotnet$S$(FSHARP_ORTOOLS_DLL_TEST)$S$(FSHARP_ORTOOLS_DLL_TEST).fsproj"
+	$(DOTNET_LIB_DIR) "$(DOTNET_EXECUTABLE)" "ortools$Sdotnet$Spackages$Sxunit.runner.console$S2.3.1$Stools$Snetcoreapp2.0$Sxunit.console.dll" "ortools$Sdotnet$S$(ORTOOLS_DLL_TEST)$Sbin$SDebug$Snetcoreapp2.0$SGoogle.$(ORTOOLS_DLL_TEST).dll" -verbose
+	$(DOTNET_LIB_DIR) "$(DOTNET_EXECUTABLE)" "ortools$Sdotnet$Spackages$Sxunit.runner.console$S2.3.1$Stools$Snetcoreapp2.0$Sxunit.console.dll" "ortools$Sdotnet$S$(FSHARP_ORTOOLS_DLL_TEST)$Sbin$SDebug$Snetcoreapp2.0$SGoogle.$(FSHARP_ORTOOLS_DLL_TEST).dll" -verbose
+
+
+.PHONY: dotnet # Build OrTools for .NET
+dotnet: \
+	clean_dotnet \
```
Mizux: not sure clean_dotnet is needed

make dotnet will always rebuild while other e.g. `make cc` will print `target 'cc' up to date nothing to do`

makefiles/Makefile.dotnet.mk
```diff
@@ -0,0 +1,251 @@
...
+	$(NUGET_EXECUTABLE) pack ortools$Sdotnet$S$(ORTOOLS_NUSPEC_FILE)
+
+.PHONY: pkg_dotnet-upload # Upload Nuget Package
+pkg_dotnet-upload: nuget_archive
+	$(warning Not Implemented)
+
+
+.PHONY: detect_dotnet # Show variables used to build dotnet OR-Tools.
+detect_dotnet:
+ifeq ($(SYSTEM),win)
```
Mizux: remove if same commands on both branches

ortools/dotnet/Google.sln
```diff
@@ -0,0 +1,76 @@
...
+		Debug|x86 = Debug|x86
+		Release|Any CPU = Release|Any CPU
+		Release|x64 = Release|x64
+		Release|x86 = Release|x86
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{AFA3F878-FB56-4314-B31E-3DA5583B548B}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
Mizux: Do we really need this magic "hash" ?

Is it univeral or dependent of your system (i.e. kind of watermarking)

acco32: This hash is the project identifier. So long as projects are added using the toolchain, this shouldn't be a problem.

Mizux: ok found ! https://msdn.microsoft.com/en-us/library/bb165951.aspx
BTW Thanks for the explanation

makefiles/Makefile.dotnet.mk
```diff
@@ -145,6 +145,11 @@ $(GEN_DIR)/com/google/ortools/sat/CpModel.g.cs: $(SRC_DIR)/ortools/sat/cp_model.
 $(GEN_DIR)/com/google/ortools/sat/SatParameters.g.cs: $(SRC_DIR)/ortools/sat/sat_parameters.proto
 	$(PROTOBUF_DIR)/bin/protoc --proto_path=$(SRC_DIR) --csharp_out=$(GEN_DIR)$Scom$Sgoogle$Sortools$Ssat --csharp_opt=file_extension=.g.cs $(SRC_DIR)$Sortools$Ssat$Ssat_parameters.proto
 
+$(CLR_KEYFILE):
+ifdef CLR_KEYFILE
```
Mizux: swap both lines ?
if CLR_KEYFILE is not def  is not necessary to create the rule `:` i.e. empty ?

acco32:
@Mizux is there another fix in master branch to fix appveyor builds.

Mizux:
Hi,

didn't know master was broken, on my way to fix it.
Seems to be related to the new `prefix` variable I have added for `make install`

Mizux:
commit https://github.com/google/or-tools/commit/df333efbcdfdd54d8988b46a2c75f524aa4d38cb
revert my patches

acco32:
@Mizux for the windows build does the `lib` folder also have to be referenced when running the application (i.e.: PATH=<ortools-root>/lib;%PATH% mono example.exe). Otherwise it seems the dotnet DLL has trouble finding the shared native library that contains the routines. Please have a look at appveyor builds.

Mizux:
@acco32 
1) on windows mono should be replaced by [`dotnet run`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-run?tabs=netcore2x) cli right ? (just confused by your i.e. example)
-> Should we also use `dotnet run` on UNIX i.e. remove mono dependencies ?
(not sure exactly if mono is(will be) still required with .Net Core 2.0 and dotnet project)

2) on windows we should **not** have a lib directory and put all .dll in bin directly since the search for dynamic library is basically the current binary directory (i.e. no /usr/lib once installed)

3) For nuget packaging I don't know where usually dll are placed inside the artifact.

4) Will appveyor be fixed if ortool.dll is placed in bin ? I means first let have a `windows style` layout (i.e. everything in bin) (note: on windows only !) then fix/adapt test and CI

mwpowellhtx:
I'm looking forward to this. Sorting out some of my bits re: ``.NET Standard`` migration as well. Will like to consider migrating my Solver adapters, as well as pick up a couple of the improved Solvers. :+1:

mwpowellhtx:
:+1:  This is awesome work. I look forward to giving it a look. Thanks!

To be clear, one doesn't "target" .NET Standard, per se. From what I understand, Standard is to Core what PCL was to portable run times from prior years. Test assemblies, applications, etc, then target 4.6.1, Core, or other actual run time, and so on. However, if I understand it correctly, virtually every other constituent assembly up until the point of commitment could potentially target .NET Standard.

I found this out somewhat to my chagrin porting one of my other suites to the .NET Standard paradigm.

Mizux:
My understanding is in theory:
.NetStandard2.0 is an API spec.  
While .NetCore2.0, .NetFramework4.6.1 and Mono5.4? are .NET implementation of these spec i.e. any .cs files containing only call(s) to these APIs must compile using one of this implementation since they implement the standard.

Now on Nuget side taking a famous package https://www.nuget.org/packages/NUnit/ ,it says 
> Supported platforms:
> - .NET Framework 2.0+
> - .NET Standard 1.6+
> - .NET Core
What's puzzle me (I'm start to getting lost) is why it is listing .Net Standard 1.6+ here, actually it seems that .NetStandard is **also** a [target platform](https://docs.microsoft.com/en-us/dotnet/standard/components#net-standard).

So inside the nunit.3.10.1.nupkg/lib you have:
* net20
* net35
* net40
* net45
* netstandard1.6
* netstandard2.0
note: netXX stand for .NetFrameworkXX
note2: looking at different packages I never saw "core2.0" directory but since .NetStandard and .NetCore are tightly linked...
open question: if you have net20 do you really need net35... ? (I think .NetFramework is backward-compatible.)

My understanding concerning runtime is `dotnet` and `mono` are runtime execution env.
- `dotnet` is cross platform but still not widely use on UNIX and rely on nuget netstandardXX dir
- `mono` is still heavily use on UNIX world (e.g. by Xamarin on MacOS) and rely on nuget netXX dir, i.e. compatible/emulate the .Net Framework...

My long term objective is to have net45 (or net46?) + netstandard2.0 dir in ortools nuget package.
And I think this is what @acco32 is targeting ?

ToDo: (in an other repo? or in a test/project directory) 
- provide one dotnet example project [using|depending on] the netstandardXX ortools nuget package dir.
- provide one mono example [using|depending on]  the netXX ortools nuget package dir.
- Integrate them in the CI job to guaranty the nuget package integrity...

mwpowellhtx:
@Mizux That's along the same lines as my loose understanding. The trouble is, if it is an "API definition" or "API surface area", whatever you call it, then why offer it as target of any assembly project at all? That was, still somewhat is, part of my confusion. "It works", but I would still like to understand it a little better. It is closer to, say, compiled byte code in Java, ubiquitous, regardless of the JVM, or the PCL, "portable" class library. Actually PCL is in the history, maturity of how we've gotten here, apparently.

Mizux:
As I said (cf link above) .Net Standard is both a "API surface area" and a "target platform" (kind of generic/universal/ubiquitous byte code that other .net implementation can consume if their are compatible with this standard).

This also explain why people ship net45 and netstandard1.6 in the nuget package since .Net Framework 4.5.1 only support NetStandard up to 1.2 cf https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support.

@all Do you think also providing `netstandard1.6` could be a good idea ? I mean there is a gap between `net45` and `netstandard2.0` but not sure it will worth the pain...

note: providing a nuget with a `net45` means `mono 5.4 >=` is only mandatory when building from source not for consuming the artifact...
many distro are still using mono 4.6/4.8 cf
[Packaging status](https://repology.org/metapackage/mono)

mwpowellhtx:
@Mizux The difference between .NET Standard and the target frameworks is not my understanding. Standard is closer to PCL than it is the run-time is my understanding. Folks ship both either because they are as royally confused about it as I am, or be cause truly their assemblies may make sense for maximum portability, and/or they also think their assemblies could stand alone, as a truly target run-time environment. One case in point is testability; test assemblies must target a run-time; **NOT** the Standard. I just got done with a round of these shenanigans, so I'm pretty well acquainted with the difference by now.

mwpowellhtx:
@Mizux To answer your broader question, don't confuse ``net45`` with ``netstandard2.0``. They are two different animals. Not sure, but probably ``netstandard1.x`` is geared towards ``net45x`` run times. Whereas, as far as I know, ``netstandard2.0`` is geared towards ``net462`` run times. I don't pretend to know why, other than, *"it is what it is"*.

Now, what that means for OR Tools... Depending what sort of features you want to leverage, could you target both? That would reach the broadest possible audience.

As for earlier mentions of fails during test assemblies? Make sure you are targeting the run times, not the API targets. That is, ``4.5.2`` and/or ``4.6.2``. That should work for you.

makefiles/Makefile.dotnet.mk
```diff
@@ -0,0 +1,251 @@
...
+CLR_ORTOOLS_DLL_NAME?=Google.$(ORTOOLS_DLL_NAME)
+BASE_CLR_ORTOOLS_DLL_NAME:=$(CLR_ORTOOLS_DLL_NAME)
+CLR_ORTOOLS_IMPORT_DLL_NAME:=$(CLR_ORTOOLS_DLL_NAME)
+
+# Check for required build tools
+ifeq ($(SYSTEM), win)
+DOTNET_EXECUTABLE := $(shell $(WHICH) dotnet.exe 2>nul)
+else # UNIX
+ifeq ($(PLATFORM),MACOSX)
+DOTNET_EXECUTABLE := $(shell dirname ${DOTNET_INSTALL_PATH})$Sdotnet
```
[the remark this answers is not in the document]
Mizux: `rehash` or `hash -r` (osx) should do the job

Mizux:
Hi,

I just merged #697 which is a deep refactoring on unix of dependencies.
I expect so trouble but I'm here to fix them ASAP !

We would also like to have this PR for the v6.8
@acco32 could you rebase your branch on master ?
Expected potential conflict
- use $L insteqd of $(LIB) for library suffix
- small fix in F# project (https://github.com/google/or-tools/commit/8c303ccac010b4fe90d506c75c24fc5bdb77807f)

acco32:
@Mizux @mwpowellhtx 
My original intent was to only target `netstandard2.0` and leave it up to the developer to compile to their target framework. I actually use this branch on another personal project targeting `netstandard`, yet I can run from my interpreter (which is running mono).  As for how far back for versioning, I can understand having `netstandard1.6` as our base, but would prefer `netstandard2.0` as it include a lot more API. The nuget package could include a separate folder for `net461` in addition to `netstandard2.0`

I will try and look this weekend to include your changes. I'd like to stay with netstandard2.0 if we can and can include a backward compatible library for netXXX (perhaps 461?). I will also include an example referencing the new library. I want to solely use the dotnet-cli and not rely on mono.

On another note, if all the files are in the same folder (rather than specify the link folder in an environment variable) then the application works as expected. When I was looking at the windows build, I didn't see the ortool.dll (referenced from Google.OrTools.dll). All the files should be packaged together when creating the nuget package.

mwpowellhtx:
@Mizux To be clear, I'm not suggesting shifting away from ``Standard v2.0``, but rather multi-target to both, if at all possible. I tend to agree. If it can target both, i.e. especially as long as the API is basically supported, I think that would reach a broader audience. Although, that being said, it's still a bit confusing to me what sort of *Long Term Support* there would be for which framework versions, from ``Standard`` to ``Core``, etc.

Re: ``CLI`` tools, makes perfect sense to me.

Re: ``ortools.dll``, I am not sure what that means. I haven't verified, personally; but, if the build is basically consistent with prior desktop framework targets if memory serves, I do not think there should be an ``ortools.dll``? In other words, as far as I know, the C++ code is built to object code and linked directly with the .NET wrapper code. But it's been a little while for me, and like I said, haven't verified re: ``Standard`` migration paths. Much of that would probably depend upon the SWIG digestion.

As for leaving it to the developer to consume into an appropriate target framework, definitely; that should work, as far as I know. ``.NET Standard v2.0`` into ``Desktop 4.6.2``, for instance, ``Standard v1.x`` into ``Desktop 4.5.2``, I think, as well, and so on.

Mizux:
@acco32  [.travis/script.sh:53](https://github.com/google/or-tools/blob/master/.travis/script.sh#L56) replace `fsharp` by `dotnet` in your branch and it should fix the [job issue on MacOS](https://travis-ci.org/google/or-tools/jobs/387610587#L2623)

acco32:
@Mizux the tests are failing due to a reference to `Google.OrTools.so` (in both osx and linux). In both platforms, `libGoogle.OrTools.so` is created. Also for osx, shouldn't the link for the dotnet wrapper be for the *dynlib* file?

Mizux:
@acco32 First thanks for fixing the build on MacOS.
Now concerning the library name I would say:
- Nice if we can remove the prefix lib (i.e. keep it for C++ stuff on unix)
- Concerning the extension, while I prefer dynlib on osx, I need to check first !
  For example for Python the `dlopen()` code only look for ".so" library even on MacOS, and since it's provided by Python authority we had to follow their borken convention (otherwise we should a resquested to everyone to build a custom python version)...
  So here, we need to know if mono/dotnet has convention/limitation when loading native libraries, then will follow them if any.

I'll try to do some search today and go back to you here, but first I need to do few changes in FSharp project...

note: if I cp lib/libGoogle.OrTools.so lib/Google.OrTools.so test pass

acco32:
I tried that earlier which didn't work. I don't think I regenerated the `*.cs` files properly with SWIG. I looked at the code again and saw the import only referenced `Google.OrTools.so`. Manually changing them to `libGoogle.OrTools.so` solved the issue. I'll have to read more about SWIG.

dotnet should have no issues loading native files through interop. In my experience, shared libraries conform to their natural platform extension and interop simply references those files. Thus `so` for linux, `dynlib` for osx, `dll` for windows would be preferred.

acco32:
Will investigate nuget packages for current targets (net462, netstandard2.0) tomorrow.

lperron:
If we have a working dotnet build (without nuget), I would push that on the
trunk.
00



Le mar. 5 juin 2018 à 12:27, Matthew Moore <notifications@github.com> a
écrit :

mwpowellhtx:
Looking forward to that NuGet, build, to be fair. But I appreciate showing
at least incremental progress.

acco32:
Found the issue with the autogenerate files not using the proper library name and made correction. On my Mac, I can reproduce the error on the build server when using the library (from test projects).

Will need to investigate the import further. As I renamed the native library to the appropriate platform (with appropriate suffix for platform), it can be read but an entrypoint cannot be found. Otherwise, the dotnet library doesn't seem to load the native dll properly. Will need to go over the include files to see if I am missing anything.

Will also have to check the tests from net462. Currently only validating against netstandard2.0 version from a netcoreapp2.0 application.

lperron:
I propose we target only netstandard2.0 on linux and mac.
00



Le mer. 6 juin 2018 à 09:03, Matthew Moore <notifications@github.com> a
écrit :

mwpowellhtx:
As long as you are adding netstandard2.0, on Windows as well. Multi-target
netstandard1.6 AND netstandard2.0, if you must still have netstandard1.6.

lperron:
I meant 4.6.2 and net standard 2.0 on windows, netstandard 2.0 only on unix.
00



Le mer. 6 juin 2018 à 14:59, Michael W Powell <notifications@github.com> a
écrit :

> As long as you are adding netstandard2.0, on Windows as well. Multi-target
> netstandard1.6 AND netstandard2.0, if you must still have netstandard1.6.
>

mwpowellhtx:
To be clear, 4.6.2 is the **RUN-TIME** that is netstandard 2.0 compatible,
i.e. for test assemblies, as is, apparently, netcoreapp2.0, for instance;
whereas netstandardx.y is an "API surface area" so-called. Really, from a
"target" perspective, M$ has really confused matters in my opinion.
netstandard is closer to PCL from yester-years. HTH

Mizux:
Well, I would say the opposite for the naming `netxxx` is the assembly\runtime compatibility and `(Net Framework) 4.6.2` is the API compatibility.

While [this page](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support) statute that `Net Framework 4.6.1` is compatible with `NetStandard 2.0` (from API point of view ?).
On the other hand it seems that `net462` only support `netstandard1.5` according to [this page](https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/#user-content-whats-new-in-net-standard-20) so targeting `net462` as "target" when building could be not equivalent to targeting `netstandard2.0` TFM.

An other concern, on unix lots of distro still use mono ~4.6 (cf above) and I don't know which Target Framework Moniker (TFM) is supported by it, clearly not `netstandard2.0`, (my bet on `netstandard1.6` or `net45`)
Otherwise started from mono 5.4 may be support `netstandard2.0` is available so if people want to use mono they have to upgrade -> in this case `netstandard2.0` is sufficient on unix. I'll try to test different configuration to verify these assertions.

Here a list of [Target Framework Moniker (TFM)](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-framework-versions)

note: on master we force mono 5.4 as minimum cf 889c97d589760ab42f4ad3540a0872f1873d11da
and we currently configure for [`net462` `netstandard2.0`](https://github.com/google/or-tools/blob/master/ortools/fsharp/OrTools.FSharp/OrTools.FSharp.fsproj#L8) but build [`net462`](https://github.com/google/or-tools/blob/master/makefiles/Makefile.fsharp.mk#L37) by default...

note2: csharp is currently built using `mcs` from `mono` on unix so I don't know which TFM it use and if it's compatible in a way with dotnet. i.e. mixing mcs build with dotnet build packages...

mwpowellhtx:
On Wed, Jun 6, 2018 at 10:14 AM, Mizux <notifications@github.com> wrote:

> Well, I would say the opposite for the naming netxxx is the
> assembly\runtime compatibility and (Net Framework) 4.6.2 is the API
> compatibility.
>
You can say the opposite, but it would be wrong. Try writing a
netstandard2.0 unit test. It doesn't work.

Mizux:
Hi,

Any news ? Should I fix the CI job for you ? I would like to merge this PR quickly then fix the remaining bug caveat if any...

mwpowellhtx:
What's the sit rep of this update? When it is done, and published mind you via NuGet, I would like to revisit my .NET wrappers, etc. Unfortunately, I don't have more bandwidth apart from helping to clarify the Standard/Core confusion the extent to which there's confusion.

Mizux:
The point is master is quickly evolving so i would like to merge this PR even if test are broken, then i could work on fixing few thing here and there. or don't hesitate to fill issue/ fixup PR.
But a lest we move forward...

acco32:
I was able to get it to compile but using the library resulted in segfaults. I'm haven't had much time to investigate. I shall look again tonight and tomorrow and then next week when I get more time.

I would prefer that the tests pass and will do the minimum to get the changes incorporated.

mwpowellhtx:
@acco32 Are these exceptions? What kind of segfaults? Provide details, I (others?) may be able to clue in as to what's going on.

acco32:
After a bit of investigation, I am able to reproduce the error from the build (unable to find native library). On my Mac, I compiled the library and tried running a test program with extra debug output. It seems the native library (libGoogle.Ortools.so) cannot load one of its dependencies:
```
Mono: DllImport attempting to load: 'libGoogle.OrTools.so'.
Mono: DllImport error loading library '.../build/libGoogle.OrTools.so': 'dlopen(.../build/libGoogle.OrTool
s.so, 9): Library not loaded: @rpath/libgflags.2.2.dylib
  Referenced from: .../build/libGoogle.OrTools.so
  Reason: image not found'.
```
I did do a make `clean_all` to remove everything and then re-compiled. @Mizux Any ideas?

mwpowellhtx:
:+1:  I am excited to updated my wrappers when this is published! Admittedly, I'm not as focused on MacOS X as I am Windows based .NET Standard/Core. But that being said, if I am targeting .NET Standard, in theory it should port.

Mizux:
Hi sorry for the delay,

Since the merge of "Use of dynamic libraries deps on UNIX" (#697)
libortools.so doesn't embed libgflags.a anymore but rely on libgflags.so.
So I had to add some rpath to [Linux](https://github.com/google/or-tools/blob/master/makefiles/Makefile.unix.mk#L156) and [MacOS](https://github.com/google/or-tools/blob/master/makefiles/Makefile.unix.mk#L216).
we have rapth:
 "." for libfoo.so to find lib.bar.so
"../lib" for bin/foo to find ../lob/libfoo.so
"../dependencies/install/lib" for bin/foo and lib/libfoo.so to find lingflags.so 

It seems that Google.ORTools.so is build inside a "build" directory accroding to your log so we should hack rpath accordingly IMHO and that should solve the link issue -> need to test your branch locally to figure out ;)

Also I think (my best bet) the compilation of `Google.ORTools.so` may not use the $(OR_TOOLS_LNK) nor $(OR_TOOLS_LDFLAGS) also.
could be checked with `otools -L Google.ORTools.so` on MacOS cf LOADER PATH meta data(s).
or "objdump -p lib/libGoogle.OrTools.so" on linux...

Also under MacOS we should have Google.ORTools.dylib I guess.
Python setuptools is borken by design (i.e. hardcoded) so even on Mac they dlopen .so we use $(SWIG_LIB_SUFFIX) I think for Google.ORTools we should use the regular $L extension.
note: I could fix it in a later patch once this one is integrated

note: you can use `make detect_cc` to inspect these values...

Mizux:
**EDIT**
*legend*: linux/mac/windows

$L is dynamic on unix but static on windows
`L = .so/.dylib/.lib` -> not a good candidate
SWIG_LIB_SUFFIX should be renamed SWIG_PYTHON_LIB_SUFFIX
`SWIG_PYTHON_LIB_SUFFIX = .so/.so/.dll` -> not a good candidate
Then create a new SWIG_DOTNET_LIB_SUFFIX with:
`SWIG_DOTNET_LIB_SUFFIX = .so/.dylib/.dll` -> obviously a good candidate
-> I'll do it once this PR is merged...
thus you should also defined:
`CLR_ORTOOLS_IMPORT_DLL_NAME:=$(LIB_PREFIX)$(CLR_ORTOOLS_DLL_NAME).$(SWIG_DOTNET_LIB_SUFFIX)`
to avoid to write $(CLR_ORTOOLS_IMPORT_DLL_NAME).$(SWIG_LIB_SUFFIX) everywhere
-> I'll do it once this PR is merged...

note: SWIG... macro are defined in **makefiles/Makefile.\[unix|win\].mk** which is loaded before `Makefile.dotnet.mk` which is important since you are using `:=` for CLR_... macro.

Mizux:
For linux I changed **OR_TOOLS_LD_FLAGS** to **OR_TOOLS_LDFLAGS** 
-> need to fix line 189 in Makefile.dotnet.mk sorry for this change I just want to follow the Makefile convention
PATCH introducing this change: https://github.com/google/or-tools/commit/6cd8da56ce4931eab664257921b1b61b29420ea2

Makefile Manual showing default rules using some default macro
CFLAGS, LDFLAGS etc so I try to follow this convention
https://www.gnu.org/software/make/manual/make.html#Catalogue-of-Rules

makefiles/Makefile.dotnet.mk
```diff
@@ -0,0 +1,277 @@
...
+	$(SED) -i -e "s/<Version>.*<\/Version>/<Version>$(OR_TOOLS_VERSION)<\/Version>/" ortools$Sdotnet$S$(FSHARP_ORTOOLS_DLL_NAME)$S$(FSHARP_ORTOOLS_DLL_NAME).fsproj
+	$(SED) -i -e "s/<AssemblyVersion>.*<\/AssemblyVersion>/<AssemblyVersion>$(OR_TOOLS_VERSION)<\/AssemblyVersion>/" ortools$Sdotnet$S$(FSHARP_ORTOOLS_DLL_NAME)$S$(FSHARP_ORTOOLS_DLL_NAME).fsproj
+	$(SED) -i -e "s/<FileVersion>.*<\/FileVersion>/<FileVersion>$(OR_TOOLS_VERSION)<\/FileVersion>/" ortools$Sdotnet$S$(FSHARP_ORTOOLS_DLL_NAME)$S$(FSHARP_ORTOOLS_DLL_NAME).fsproj
+	"$(DOTNET_EXECUTABLE)" build -c Release -o $(BIN_DIR) ortools$Sdotnet$S$(ORTOOLS_DLL_NAME)$S$(ORTOOLS_DLL_NAME).csproj
+	"$(DOTNET_EXECUTABLE)" build -c Release -o $(BIN_DIR) ortools$Sdotnet$S$(FSHARP_ORTOOLS_DLL_NAME)$S$(FSHARP_ORTOOLS_DLL_NAME).fsproj
+	$(MKDIR_P) $(DOTNET_BUILD_DIR)
+	"$(DOTNET_EXECUTABLE)" publish -c Release -o "..$S..$S..$S$(DOTNET_BUILD_DIR)" -f netstandard2.0 ortools$Sdotnet$S$(FSHARP_ORTOOLS_DLL_NAME)$S$(FSHARP_ORTOOLS_DLL_NAME).fsproj
+
+
+BUILT_LANGUAGES +=, dotnet (netstandard2.0)
```
Mizux: $ make clean
/bin/sh: -c: line 0: syntax error near unexpected token `('
/bin/sh: -c: line 0: `echo Or-tools have been cleaned for C++ , Python2.7 , Java , dotnet (netstandard2.0)'
Makefile:97: recipe for target 'clean_all' failed
make: *** [clean_all] Error 1

-> I think echo doesn't like your parentheses...

Mizux: Done

Mizux:
Seems (on linux) that your copy of all dependencies/install/lib/* is not necessary anymore...
-> removing it...

Mizux:
Appveyor build fails due to random.cc file
lperron has fixed it https://github.com/google/or-tools/commit/8bc9b159598854cc80434cb09a6b92c7cf040ddd

I'll try to rebase acco32/or-tools#dotnet:
```sh
$ git clone git@github.com:acco32/or-tools.git acco32
$ cd acco32
$ git checkout dotnet
$ git remote add google git@github.com:google/or-tools.git
$ git fetch --all
$ git rebase google/master
having fun with meld !
```

Mizux:
@acco32 I just rebased and `push -f` your branch, hope you won't blame me for this ;)

Mizux:
https://github.com/google/or-tools/commit/ae79a832ccb0492507cb6456d3055dc6d75f711b fix master windows -> need to rebase acco32/dotnet on it...

Mizux:
issue in clean_dotnet, on windows command separator is `&` not `;`.
On my way to fix it properly and test in my windows VM please wait...

src:
```cmd
tools\rm.exe -rf ortools\\gen
tools\rm.exe -rf .\\dotnet-test
del bin\\Google.Protobuf.*;  del bin\\Google.OrTools.*;  del bin\\Google.OrTools.*;
Could Not Find C:\projects\or-tools-98u1n\bin\Google.Protobuf.*
Could Not Find C:\projects\or-tools-98u1n\del
Could Not Find C:\projects\or-tools-98u1n\bin\Google.OrTools.*
Could Not Find C:\projects\or-tools-98u1n\del
Could Not Find C:\projects\or-tools-98u1n\bin\Google.OrTools.*
sn -k bin\\or-tools.snk
```

Mizux:
Ok currently Windows will always failed because:
`$(BIN_DIR)/$(CLR_ORTOOLS_DLL_NAME)$(DLL)` which resolve in `bin/Google.OrTools.dll`
and it depends on `$(LIB_DIR)$S$(LIB_PREFIX)$(CLR_ORTOOLS_DLL_NAME).$(SWIG_LIB_SUFFIX)`
which resolve in `lib/Google.OrTools.dll` (on unix `lib/libGoogle.OrTools.so`)

On linux/mac we use `(DY)LD_LIBRARY_PATH=lib dotnet....` to test, while on windows we copy `lib/*` and `bin/*` in `dotnet-test` -> one binary and one library with the same name -> everything blow up :boom: :fireworks: :
Trace on myWindows VM
```
"C:\Program Files\dotnet\dotnet.exe" \
 "ortools\\dotnet\\packages\\xunit.runner.console\\2.3.1\\tools\\netcoreapp2.0\\xunit.console.dll" \
 ".\\dotnet-test\\Google.OrTools.Tests.dll" \
 ".\\dotnet-test\\Google.OrTools.FSharp.Tests.dll" -verbose
xUnit.net Console Runner (64-bit .NET Core 4.6.26328.01)
  Discovering: Google.OrTools.Tests
  Discovered:  Google.OrTools.Tests
  Starting:    Google.OrTools.Tests
    Google.OrTools.Tests.LinearProgramming.TestBinaryOperations
    Google.OrTools.Tests.LinearProgramming.TestBinaryOperations [FAIL]
      System.TypeInitializationException : The type initializer for 'Google.OrTools.LinearSolver.Solver' threw an exception.
      ---- System.TypeInitializationException : The type initializer for 'Google.OrTools.LinearSolver.operations_research_linear_solverPINVOKE' threw an exception.
      -------- System.TypeInitializationException : The type initializer for 'SWIGExceptionHelper' threw an exception.
      ------------ System.EntryPointNotFoundException : Unable to find an entry point named '?' in DLL 'Google.OrTools.dll'.
      Stack Trace:
        C:\Users\corentinl\source\repos\acco32\ortools\dotnet\OrTools.Tests\LinearProgramming.cs(101,0): at Google.OrTools.Tests.LinearProgramming.TestBinaryOperations()
        ----- Inner Stack Trace -----
           at Google.OrTools.LinearSolver.operations_research_linear_solverPINVOKE.Solver_CLP_LINEAR_PROGRAMMING_get()
        C:\Users\corentinl\source\repos\acco32\ortools\gen\com\google\ortools\linearsolver\Solver.cs(214,0): at Google.OrTools.LinearSolver.Solver..cctor()
        ----- Inner Stack Trace -----
           at Google.OrTools.LinearSolver.operations_research_linear_solverPINVOKE.SWIGExceptionHelper..ctor()
        C:\Users\corentinl\source\repos\acco32\ortools\gen\com\google\ortools\linearsolver\operations_research_linear_solverPINVOKE.cs(125,0): at Google.OrTools.LinearSolver.operations_research_linear_solverPINVOKE..cctor()
        ----- Inner Stack Trace -----
           at Google.OrTools.LinearSolver.operations_research_linear_solverPINVOKE.SWIGExceptionHelper.SWIGRegisterExceptionCallbacks_operations_research_linear_solver(ExceptionDelegate applicationDelegate, ExceptionDelegate arithmeticDelegate, ExceptionDelegate divideByZeroDelegate, ExceptionDelegate indexOutOfRangeDelegate, ExceptionDelegate invalidCastDelegate, ExceptionDelegate invalidOperationDelegate, ExceptionDelegate ioDelegate, ExceptionDelegate nullReferenceDelegate, ExceptionDelegate outOfMemoryDelegate, ExceptionDelegate overflowDelegate, ExceptionDelegate systemExceptionDelegate)
        C:\Users\corentinl\source\repos\acco32\ortools\gen\com\google\ortools\linearsolver\operations_research_linear_solverPINVOKE.cs(105,0): at Google.OrTools.LinearSolver.operations_research_linear_solverPINVOKE.SWIGExceptionHelper..cctor()
...
```

Since windows doesn't have env variable we need to copy so we need a different name e.g. `LIB_PREFIX=lib` on windows....

@acco32 I think I'll fork your branch acco32/dotnet until finding something working then cherry pick it in your branch.... or `push -f` on your branch until it works :japanese_ogre:

mwpowellhtx:
@Mizux There should be sufficient separation between platforms that if you need to do something Windows specific, if necessary, you should be able to do that. In other words, if it isn't behaving like Mac or Linux, that's by design; don't try to force it to fit through the round hole.

Mizux:
@mwpowellhtx  Yeah granted it's just on Windows we don't use "lib" prefix and dynamic lib extension are 'dll" like dotnet library -> I'm just fighting against our own naming convention to keep consistency. Just feel odd to have libGoogle.OrTools.dll (aka libFoo.dll) on windows while all others lib on windows are called foo.lib or foo.dll.

Mizux:
@mwpowellhtx 
> You can say the opposite, but it would be wrong. Try writing a
netstandard2.0 unit test. It doesn't work.

In fact it doesn't work because netstandard2.0 is **not** a *runnable* TFM.
so you can target it when building a [library](https://github.com/dotnet/samples/blob/master/core/getting-started/unit-testing-using-dotnet-test/PrimeService/PrimeService.csproj#L4) not and [executable](https://github.com/dotnet/samples/blob/master/core/getting-started/unit-testing-using-dotnet-test/PrimeService.Tests/PrimeService.Tests.csproj#L4) (aka `<OutputType>Exe</OutputType>`)

doc: https://github.com/dotnet/docs/blob/master/docs/standard/frameworks.md
related issue: https://github.com/dotnet/sdk/issues/428

mwpowellhtx:
@Mizux Yes, exactly. That's what I was saying. It is a *target*, but it is *not* a *"runtime"*. If I was reading correctly, there seemed to be some confusion over what should be targeted where throughout this update.

acco32:
Hey all,

Quite a lot of activity going on. @Mizux, thanks for taking a look and making the necessary changes for the wibdows build.

- have a look at the readme I included to see the flow I envisioned about why the test/build dependencies are the way they are. By issuing make test, I only rebuild the main lib and the test project.

- the build folder is the next step for the nuget package at thw command `dotnet publish` collects everything to be deployed. the bin folder can still be used to collect the final product.

- does the name of the windows native library really matter if for dotnet they'll always reference `Google.OrTools.dll`? Perhaps we should rename the dotnet lib (e.g. `Google.OrTools.dotnet.dll`) and leave everything as is (comment is too late probably)

- the dotnet build should be platform independent with the exception of the autogenerated pinvoke files (for the library name).

Anyway, thanks for looking and we probably want to close off this branch and start a new once windows compiles. 

What is the next component on your roadmap for the dotnet client? Has anyone looked at updating the interface for the existing C# client?

mwpowellhtx:
@acco32 My two cents.

> does the name of the windows native library really matter if for dotnet they'll always reference Google.OrTools.dll? Perhaps we should rename the dotnet lib (e.g. Google.OrTools.dotnet.dll) and leave everything as is (comment is too late probably)

IMO, no; leave the names as-is. What will need to change is the packaging for NuGet, however; with the targeting for .NET Standard 2.0. NuGet should pick up these bits automatically and inform you one way or the other what else needs to happen from the consumer side of things.

> the dotnet build should be platform independent with the exception of the autogenerated pinvoke files (for the library name).

Not sure what you mean "exception of autogenerated"... Targeting ``netstandard2.0`` is targeting ``ns2.0``, and so on. As long as the API surface area supports those calls, all should be seamless, AFAIK.

> What is the next component on your roadmap for the dotnet client? Has anyone looked at updating the interface for the existing C# client?

IMO, Nuget packaging and publication needs to be a clearly identified milestone on that roadmap whatever else anyone has in mind. For instance, what do you mean, *"updating the interface"*?

Mizux:
:champagne: Appveyor CI i.e. windows pass the compilation **and** test.

One thing I will change is how test are run, currently we have:
```
LD_LIBRARY_PATH="lib:" "/usr/local/google/home/corentinl/dotnet/dotnet" \
 "ortools/dotnet/packages/xunit.runner.console/2.3.1/tools/netcoreapp2.0/xunit.console.dll" \
 "./temp_dotnet_test/Google.OrTools.Tests.dll" \
 "./temp_dotnet_test/Google.OrTools.FSharp.Tests.dll" -verbose
```
I would like to be able to use:
```
dotnet test -o "../../../temp_dotnet_test"  ortools/dotnet/OrTools.Tests
```

mwpowellhtx:
@Mizux It should [support xunit tests](http://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test), but I'm not sure what the sitrep is concerning results reporting.

Mizux:
@mwpowellhtx  wrong runner xunit.console is not working with dotnet...

mwpowellhtx:
@Mizux The Microsoft docs [claim support for xunit](http://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test), but I cannot promise what other details need sorting out. You may also look into [``dotnet xunit`` as contrasted with ``dotnet test``](/xunit/xunit/issues/1407).

Mizux:
I mean if I replace xunit.runner.visual by xunit.console it doesn't work.

I prefer our project follow Microsoft documentation...

Also calling `dotnet ....xunit...dll OrTools.Tests.dll` is not documented
in dotnet help..

...

Le jeu. 28 juin 2018 à 18:37, Michael W Powell <notifications@github.com> a
écrit :

mwpowellhtx:
@Mizux I don't know what to tell you. I don't have the bandwidth to dive into this thing very deep, apart from what the docs, issues, etc, appear to indicate one way or another. Sometimes preferences need to change when faced with reality on the ground.

Mizux:
master broken;  https://github.com/google/or-tools/commit/94b539b77537f9441ce5acaa64f68431bdb645a7 should fix it.
if ok I'll rebase the branch on it -> wait and see

Mizux:
1:26am here see you tomorrow...
@all if everything pass I'll merge it tomorrow...
Travis CI doesn't appear because already build in mizux/windows cf https://travis-ci.org/google/or-tools/builds/397997526

note for myself:
- ￼CLR_ORTOOLS_FSHARP_DLL_NAME should depends on ￼CLR_ORTOOLS_DLL_NAME not ortoolslibs.
- why a second `make dotnet` rebuild some stuff
- ortools version is changed inline in *.\[cf\]sproj, let's change it 
- need to check the dotnet sln
- dotnet readme to rewrite
- nugets package to do

acco32:
@Mizux thanks for the extra work on the makefile format (make it match the standard the rest of OR-Tools is using). I do have some questions regarding the structure but perhaps we can take those offline in a different format. Thanks again.

mwpowellhtx:
@acco32 @Mizux What's the sitrep here? Any progress? Looking forward to seeing that NuGet package published. Thanks!

Mizux:
Hi all,

@acco32 @mwpowellhtx I continue some refactoring on dotnet which currently live on master...
I also take few bits of #744 (Nugget package) but still didn't test it.
Hope to do nuget test this week (second half).

Currently I'm verifying our docker distro can build master a create working archive once done I'll test the Nuget generation...

mwpowellhtx:
@Mizux I'm really interested to take a look at at least a solid prerelease RC package right now. If confidence is high this is on a winning track, then perhaps examples could be packaged separately from now on, i.e. ``google.ortools.examples.$(semver).nupkg``, and subsequently delivered, with appropriate dependencies on the ``google.ortools.$(semver).nupkg``. Thanks!

mwpowellhtx:
[Issue filed](/google/or-tools/issues/760) [along these lines](#issuecomment-405290218). If it can be easily addressed, awesome! Without distracting from the overall goal of this endeavor, that is.

paultrow:
The installations for OR-Tools with .Net currently say the following:

"Prerequisite: .NET Core 2.1 SDK
Then you must also install the package .NET Core 2.1 SDK version 2.1.302 or higher:

Simply install this https://www.microsoft.com/net/download/thank-you/dotnet-sdk-2.1.302-windows-x64-installer"

Does support for .NET Standard 2.0  mean that users could install .NET Standard instead of .NET Core? Would this be of interest to the ordinary OR-Tools user, who is not building applications?

Thanks.

mwpowellhtx:
@paultrow I do not know about OR Tools docs, but to be clear, as I understand it and has been explained to me, Standard is not Core, and Core is not Standard. Standard is an "API surface area" so to speak, whereas Core is a run-time. Standard dependencies are closer to PCL from previous years, and can be consumes by, i.e. are dependencies for, Core projects, .NET Framework 4.6, 4.7, depending on the Standard version, and so on. I don't know if this answers your question, but it's not as simple as *"depending on Core"*.

mwpowellhtx:
@lperron What's the situation re: this effort? Where does the test code stand? Let's separate those packages so we can see some measurable progress publishing at least an RC package for the OR Tools themselves, shall we? Thanks!

paultrow:
@mwpowellhtx My main question is who among OR-Tools users is support for .NET Standard relevant to?  I'm guessing that for the vast majority of users, it isn't important.

paultrow:
I apologize if I offended you. I'm simply gathering information for the OR-Tools Release Notes and installation instructions. I'll look elsewhere.

mwpowellhtx:
@paultrow I don't know what your background is, your Github profile seems rather barren at the moment, for instance. Core/Standard is/was/continues to be confusing even for the "washed" so to speak, so the confusion is understandable.

Personally, I've made an effort to get informed about it; for instance, .NET Foundation holds a weekly (for the most part) YouTube/Hangouts stand up that covers the latest goings on concerning ASP.NET Code, Standard, and Core timelines, etc. Gitter is also an interesting source to learn about discussions of this nature.

I'm not sure what is meant by "ordinary OR Tools users" or "who is not building applications"? In what sense of the word "applications"? Desktop, web-based, component assemblies? At least from a .NET perspective, OR Tools is not an application unto itself, at least not as far as I am concerned.

lperron:
Oh, let's calm down :-)

Paul is our technical writer working on or-tools.

I decided we would offer a nuget package build with dotnet core 2.0.
From what I understand, it can be used by dotnet core directly (useful for
non windows users), or consumed by visual studio (latest vs2015 or 2017 if
I am not wrong).

We released a stable version in 6.8 that included lot of unbundling,
especially around all third parties we depend on.

Building the nuget package did not make the cut. but should be available in
August with v 6.9.

Did I miss something?




00

jonwagner:
".NET Standard 2.0" is basically a standard set of library APIs that you code to. If or-tools targets .NET Standard, then you can run it on any platform that supports the standard.

".NET Core 2.0" is a similar API, but is intended for end applications. That should be out of the discussion here.

If you want to keep "legacy" code alive, you will also likely want to ship a NET45 or NET46 target. There are plenty of shops that are stuck on NET4x and won't be updating in the near future.

Personally, I try to target all new .NET code to the standard, so you have the option of running portably (e.g. I'm trying to run on OSX and Win for this project).

TL;DR - target `netstandard20` and (`net45` or `net46`)

I'd be happy to help out with this bit if you need some additional resources.

lperron:
Question,

1) can we build all of these with net core 2.1 SDK?
 Hopefully yes,

2) How? All config files are in ortools/dotnet/

Thanks
00



Le lun. 30 juil. 2018 à 09:24, Jon Wagner <notifications@github.com> a
écrit :

jonwagner:
You can use the same tools to build any of those targets, but if you want to compile a net45 target, I think you have to do the build on a windows machine. Since you're producing a nuget package already, your build system should support it.

I'm new to your code. I'll take a deeper look at it tonight (TZ = America/New_York)

mwpowellhtx:
@lperron @jonwagner I just want to be clear on the language here. Understandably, OR Tools does not exclusively serve the .NET community, nor even necessarily Windows only, nor should it. That's where folks like myself and @jonwagner enter in.

Note, see also [Target frameworks](http://docs.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-framework-versions).

Yes, netstandard is the "API surface area", akin to PCL, which is what you want, I should think. netcore is a run time, in theory extending across platforms. However, pretty sure both "targets" extend across platforms. Depending on the netstandard version, also supports net46 and net47 depending on whether you target netstandard1.x or netstandard 2.x, but I could be wrong there, net45 may work here as well, I don't know.

Similarly with @jonwagner, I've accepted that netstandard is the appropriate target for many of my sort of basic, building block assemblies. Core may be **A** runtime, does not need to be **THE** target that I adopt; I may want to target net47, depending on the requirement, for instance. The point is, when I consume a netstandard package, then I am assuming PCL-like flexibility until the very last moment when I need to target a run-time.

Tracking this, or other issues, however, I've had the impression that the samples, etc, as been the bottle neck to packaging anything, never mind an RC. The sooner packaging can happen the sooner I can start updating my own wrappers when the dependencies are known. So anything we can do to clear the path for the "core" OR Tools assembly, if you will, I'm interested to see happen. This is my interest in it as you know.

Others have been forging this path for now, so I am reporting as an interested bystander, but I can provide feedback, contribute further, etc, provided bandwidth is available.

lperron:
Thanks.

As seen in ortools/dotnet/OrTools/OrTools.csproj.in, we are currently
targeting netstandard 2.0.

So you think that on windows we should also target 4.x? Would 4.5 be the
right target?

Thanks
00

lperron:
Or as seen here:

https://docs.microsoft.com/en-us/dotnet/standard/net-standard

netstandard2.0 and NET Framework 4.6.1 are the same, and both should work
OK.

I am just trying to do the minimum amount of relevant work. For instance,
for non windows, I have dropped mono in favor of .NET Core 2.1 as the build
tool.

Jon, I am open to all suggestions, pull requests to build the right nuget
package.
The immediate goal is a windows nuget.

Then we can look at building a cross platform nuget package. When this is
done, this will be the only binary channel for releasing OR-Tools on .NET.

Thanks for all your advices.


00

mwpowellhtx:
@lperron Thanks for clarifying by the way; I was concerned we might be talking about targets themselves, but I probably read that incorrectly. We're on the same page; if you can target one, or a couple at most, instead of half a dozen, absolutely.

> So you think that on windows we should also target 4.x? Would 4.5 be the right target?

Assuming any sort of backwards compatibility is required, then it sounds like the right thing to do. I'd be content to leave it there as a dormant issue, and if anyone still needed it to up vote it, or give it a thumbs up, or something along these lines.

For my concerns, I am more interested in tracking with the netstandard path these days, however. Maybe there is a corner case where I still need the ``net45`` support, but I cannot think of one off the cuff.

jonwagner:
I was only suggesting a .NET Framework target to not break existing users.

The last version only targets `Google.OrTools 6.7.4973' was restored using '.NETFramework,Version=v4.6.1'`, so keeping 4.6.1 support should keep people happy.

If you build for netstandard20, it *may* work on 461, but may not (see https://github.com/dotnet/standard/issues/64 for some notes)

Definitely build for netstandard2.0. It will get you everything going forward.

If your build process already runs on a Windows box, you might be able to target both for (almost) free just by specifying both targets in the .csproj file. (example: https://github.com/jonwagner/Insight.Database/blob/master/Insight.Database.Core/Insight.Database.Core.csproj)

IMO: if you get it for free, keep net461 and netstandard20. If you have to choose, go netstandard20.

I'll check out the code later...

mwpowellhtx:
@lperron @jonwagner Good points.

My understanding is that ``netstandard1.x`` targets are ``net46x`` compatible, whereas ``netstandard2.x`` is ``net47x`` compatible moving forward.

Personally, I'm less interested in anything but ``netstandard2.x`` at this point, but like I said, leaving it open for others to up vote, etc, may be a good thing.

Anywho... Thanks, all!

mwpowellhtx:
@paultrow @lperron @jonwagner One thing to also keep in mind also is whether you want to be on LTS. I'm not positive if that has to do with the SDK, per se; I'm a bit fuzzy on those details. As contrasted with a daily or nightly Core/SDK builds, let's say.

lperron:
normally, from sources,
install or-tools, install dependencies :
https://developers.google.com/optimization/install/dotnet/source_mac
Note: when we target 2.0, we do not need to install mono.
To target 1.6, I believe we need mono.

Then make dotnet
make test_dotnet

My idea was to target only netstandard 2.0 on unix, and maybe 1.6 and 2.0
on windows.

--Laurent
00



Le lun. 30 juil. 2018 à 19:19, Jon Wagner <notifications@github.com> a
écrit :

mwpowellhtx:
@lperron I couldn't advise you as to the Linux dependencies, but as far as the ``netstandard1.6`` (or greater), that likely may extend across platforms. So, if I understand @jonwagner 's point, as long as the API surface area provides sufficient coverage, it should be fine for that or better ``netstandard`` versions across platforms. Test all this, of course, but this is my loose understanding. Notwithstanding *Mono* dependencies. Personally, I'm fine just saying ``netstandard2.0`` or better, but I'm mainly focused in the Windows area at the moment; but the API coverage makes sense to me.

jonwagner:
I spent some time trying to get or-tools to compile against `netstandard1.6`. SWIG uses some classes that aren't available, and it would be non-trivial to change it. So my recommendation is to target `netstandard2.0` like you currently are.

I'll be building that and I can give you feedback on how the packaging looks.

lperron:
Yes, PInvoke stuff requires 2.0. I forgot.

2.0 it is then..
Thanks.
00

Mizux:
To summary: 
* OrTools target `netstandard2.0` as a library to only depends on Microsoft SDK (@paultrow the SDK contains the netstandard2.0 libraries and netcore2 runtime) and remove deps to `mono` on unix (i.e. "loosing" netfx framework in favor of [Microsoft backed netstandard](https://www.microsoft.com/net/learn/get-started-with-dotnet-tutorial)), while 
* examples target `netcore2.0` as runtime "last end product" (also available in the SDK).
* to get rid of mono we generate the ortools.snk key file using a dotnet application cf [tools/dotnet/CreateSigningKey](https://github.com/google/or-tools/tree/master/tools/dotnet/CreateSigningKey) unfortunatelly all sdk were [broken](https://github.com/dotnet/cli/pull/9330) on unix to correctly deal with relative path and [`GetCurrentDirectory()`](https://github.com/google/or-tools/blob/master/tools/dotnet/CreateSigningKey/Program.cs#L12) stuff so we have to target `netcore2.1` and SDK 2.1.302 (note: this only concern CreateSigningKey, examples can target `netcore2.0` since they are not using this API)

* All previous examples have been kept in examples/dotnet/netfx and are not built since they depends on netfx which is not available on unix unless using mono -> need to make my mind about them...
* I'll try to move all examples to `netcore2.0` and reference OrTools `netstandard2.0`.
* Currently nuget package generation is broken please see #744 to follow.
As @lperron said first we'll try to provide a window only native package then try to provide a cross platform one next.

ps: in vacation last week sorry for the delay

PR #680, opened 2018-05-03
Fix comments of the SequenceVar Rank functions

alanbato:
Make comments regarding RankLast and RankNotLast coherent, they had the RankFirst and RankNotFirst descriptions.

I know it's a small change, but I've seen a few of this mismatched comments and decided I should give it a go and do something about it :)

I can't recall which others had the same error as these ones, but I'll be on the lookout for them.


lperron:
Thanks

PR #692, opened 2018-05-15
port wedding_optimal_chart.cs to Python

turadg:
This change is [Reviewable](https://reviewable.io/reviews/google/or-tools/692)


lperron:
Hi, 

Thanks.
I have pushed the same example, but using the CP/SAT solver.

PR #697, opened 2018-05-23
Use dynamic dependencies on UNIX

Mizux:
Use dynamic dependencies on unix instead of static libs.


makefiles/Makefile.python.mk
```diff
@@ -377,104 +508,206 @@ else
...
+$(PYPI_ARCHIVE_TEMP_DIR)/ortools/README.txt: tools/README.pypi | $(PYPI_ARCHIVE_TEMP_DIR)/ortools
+	$(COPY) tools$SREADME.pypi $(PYPI_ARCHIVE_TEMP_DIR)$Sortools$SREADME.txt
+
+$(PYPI_ARCHIVE_TEMP_DIR)/ortools/LICENSE-2.0.txt: LICENSE-2.0.txt | $(PYPI_ARCHIVE_TEMP_DIR)/ortools
+	$(COPY) LICENSE-2.0.txt $(PYPI_ARCHIVE_TEMP_DIR)$Sortools
+
+PYTHON_SETUP_DEPS=
+ifeq ($(UNIX_GFLAGS_DIR),$(OR_TOOLS_TOP)/dependencies/install)
+  ifeq ($(PLATFORM),MACOSX)
+    PYTHON_SETUP_DEPS += , 'libgflags.2.2.$L'
```
Mizux: Should be $(PYTHON_SWIG_SUFFIX) on MacOS i.e. python look for .so even on Mac -_-

Mizux: it's ok this is dynamic dependencies not pywrappers so on MacOS they are suffixed by .dylib

Mizux:
I do my best to rewrite all rpath and carefully ship all dependencies in the bundle so python user should be not impacted i.e. will still ship standalone python (ed: it also requires six and protobuf python package since v6.6 if I remember well).

Please note that now we use CBC from github mirror which is splitted in five repo (CoinUtils, Osi, Clp, Cgl, Cbc). We also do this since most distro ship Cbc like this (and coin-or.org is regularly down...)

The ultimate goal is to prepare or-tools to be able to build and package it without its dependencies so we could integrate libortools to any linux/brew distro as a regular package which depends on other libraries...
note: Experimental homebrew package here https://github.com/Mizux/homebrew-or-tools (only --Head is working) I'll try to push it to homebrew-core once we release the v6.8


Feedback or question are welcome !

Issue #699, opened 2018-05-24
glog FATAL level logging exits process and causes Python program to exit as well


yingzong:
There are parts in the ortools library where FATAL severity errors are logged through the glog library. It turns out glog forces the calling process to exit after logging, on such errors, see:
http://rpg.ifi.uzh.ch/docs/glog.html

This is a fairly annoying pattern, especially for a library like ortools that is expected to be called by others (including Python programs). Many times the error is something like network connection failure or something that should be retried and is recoverable. I think there won't be an easy fix here but have there been thoughts on how to address this? Ideally even fatal severity errors raise exceptions so that they can be handled differently depending on the calling program (the default can be to exit, of course).

N.B.: There is also a DFATAL severity that only exits the program in debug mode but converts to ERROR severity in production. Might that be appropriate?

lperron:
PR was merged.
Closing the issue.

PR #700, opened 2018-05-25
Add exception handling to the solver creation call

yingzong:
This is a very targeted addition, which can be generalized. It
makes the following changes:

1. In the solver creation call of the Gurobi interface, i.e.
the constructor `GurobiInterface::GurobiInterface`, the GLOG
severity level is reduced to `DFATAL`, which becomes `ERROR` in
production. This means errors in calls to `GRBloadenv` no longer
automatically kill the process as a side effect.

2. An `std::runtime_error` exception is raised with the return code
and the error message from GRBloadenv. This replaces the functionality
of exiting the process when this error is unhandled. Additionally,
it enables alternate handling by the caller.

3. In the linear solver SWIG wrapper, exception wrapping is turned
on specifically for the `MPSolver::MPSolver` constructor call and
the `std::runtime_error` that may now occur there. This exception
is re-raised as the corresponding RuntimeError in the target language.

Addresses issue #699.


yingzong:
@Mizux Any thoughts on the likelihood of this being merged?

PR #741, opened 2018-07-04
Add examples for C#

acco32:


examples/dotnet/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+We recommend that all projects you create target `netcoreapp2.0` as this allows you to compile for various frameworks and keep up-to-date with the latest frameworks.
+
+Wherever you have ortools installed, be sure to reference the `Google.OrTools.dll` from the project file. You will also need to reference the library folder housing native libraries. 
+
+### Linux
+To reference a particular folder on linux, you can either: explicitly set the **LD_LIBRARY_PATH**; or create a new configuration file with the path of the library folder in `/etc/ld.so.conf.d/` and then run `sudo ldconfig`. The former will set the path on a system level so that you don't have to use the environment.
+
+### MacOS
+To reference a particular folder on linux, you can explicitly set the **DYLD_FALLBACK_LIBRARY_PATH**
```
Mizux: prefer to use DYLD_LIBRARY_PATH it seems the fallback one doesn't play well with homebrew and security.

src: https://github.com/boxen/puppet-java/pull/37

Mizux:
@acco32  Just a question, we already have examples in examples/csharp and examples/fsharp in this case why don't using `git mv` ?

acco32:
We can. I want to create projects referencing net core (Both C# and F#) as well as organize the folder a bit better (i.e.: add remove examples as necessary with some better instructions)

examples/dotnet/OrTools.sln
```diff
@@ -3,7 +3,9 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.26124.0
 MinimumVisualStudioVersion = 15.0.26124.0
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "examples", "csharp\examples.csproj", "{0899C5EB-2AD1-49C1-9AB3-735E5B81BF56}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "csharp", "csharp\csharp.csproj", "{0899C5EB-2AD1-49C1-9AB3-735E5B81BF56}"
+EndProject
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "examples", "fsharp\fsharp.fsproj", "{B0E80F2F-BDB3-4688-8E27-5678173DC28C}"
```
Mizux: let's call it "fsharp"

examples/dotnet/README.md
```diff
@@ -10,12 +10,12 @@ Wherever you have ortools installed, be sure to reference the `Google.OrTools.dl
 To reference a particular folder on linux, you can either: explicitly set the **LD_LIBRARY_PATH**; or create a new configuration file with the path of the library folder in `/etc/ld.so.conf.d/` and then run `sudo ldconfig`. The former will set the path on a system level so that you don't have to use the environment.
 
 ### MacOS
-To reference a particular folder on linux, you can explicitly set the **DYLD_FALLBACK_LIBRARY_PATH**
+To reference a particular folder on linux, you can explicitly set the **DYLD_LIBRARY_PATH**
 
-## CSharp
+## CSharp/FSharp project examples
 
 By default all the examples are compiled in a console applicaiton with the startup object being the **Classname.Main** so that when compiled the entrypoint will be known.
```
Mizux: application

Mizux: application

Mizux:
Two questions:
Did you remove some examples or just move them ?

Why did you put them in csharp-netfx do we rely on netframework ? I believe we only use netstandard2 and netcore2 ?

Sorry for this dumb questions

acco32:
I moved them. The library itself is made in netstandard2.0. We don't rely on the framework any more but I left them in to show an example of compiling with the framework if that is all you have. dotnet core right now does not allow you to compile single source files. For instance the F# examples use the library with a netstandard reference yet they are compiled with mono. The other examples reference the netstandard library but are built with netcore.

Mizux:
if OK, let's merge it as it.

PR #744, opened 2018-07-06
Create nuget package (without native libs)

acco32:
@Mizux I am unsure how you plan to package the native libraries for each operating system. This target will create the package for dotnet only. Once the native library installation is decided we can add it as a dependency to this package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,8 +65,6 @@ include $(OR_ROOT)makefiles/Makefile.cpp.mk
 include $(OR_ROOT)makefiles/Makefile.python.mk
 include $(OR_ROOT)makefiles/Makefile.java.mk
 include $(OR_ROOT)makefiles/Makefile.dotnet.mk
-# include $(OR_ROOT)makefiles/Makefile.csharp.mk
-# include $(OR_ROOT)makefiles/Makefile.fsharp.mk
 include $(OR_ROOT)makefiles/Makefile.archive.mk
 include $(OR_ROOT)makefiles/Makefile.install.mk
 
--- a/makefiles/Makefile.dotnet.mk
+++ b/makefiles/Makefile.dotnet.mk
@@ -368,9 +368,6 @@ dotnet: ortoolslibs csharp_dotnet fsharp_dotnet
 	$(SED) -i -e "s/<FileVersion>.*<\/FileVersion>/<FileVersion>$(OR_TOOLS_VERSION)<\/FileVersion>/" ortools$Sdotnet$S$(ORTOOLS_FSHARP_DLL_NAME)$S$(ORTOOLS_FSHARP_DLL_NAME).fsproj
 	"$(DOTNET_EXECUTABLE)" build -c Release -o $(BIN_DIR) ortools$Sdotnet$S$(ORTOOLS_DLL_NAME)$S$(ORTOOLS_DLL_NAME).csproj
 	"$(DOTNET_EXECUTABLE)" build -c Release -o $(BIN_DIR) ortools$Sdotnet$S$(ORTOOLS_FSHARP_DLL_NAME)$S$(ORTOOLS_FSHARP_DLL_NAME).fsproj
-	$(MKDIR_P) $(TEMP_DOTNET_DIR)
-	"$(DOTNET_EXECUTABLE)" publish -c Release -o "..$S..$S..$S$(TEMP_DOTNET_DIR)" -f netstandard2.0 ortools$Sdotnet$S$(ORTOOLS_DLL_NAME)$S$(ORTOOLS_DLL_NAME).csproj
-	"$(DOTNET_EXECUTABLE)" publish -c Release -o "..$S..$S..$S$(TEMP_DOTNET_DIR)" -f netstandard2.0 ortools$Sdotnet$S$(ORTOOLS_FSHARP_DLL_NAME)$S$(ORTOOLS_FSHARP_DLL_NAME).fsproj
 
 BUILT_LANGUAGES +=, dotnet \(netstandard2.0\)
 
@@ -385,7 +382,12 @@ NUGET_SRC = https://www.nuget.org/api/v2/package
 
 .PHONY: pkg_dotnet # Build Nuget Package
 pkg_dotnet:
-	$(warning Not Implemented)
+	$(MKDIR_P) $(TEMP_DOTNET_DIR)
+	"$(DOTNET_EXECUTABLE)" publish -c Release --no-dependencies --no-restore -o "..$S..$S..$S$(TEMP_DOTNET_DIR)" -f netstandard2.0 ortools$Sdotnet$S$(ORTOOLS_DLL_NAME)$S$(ORTOOLS_DLL_NAME).csproj
+	"$(DOTNET_EXECUTABLE)" publish -c Release --no-dependencies --no-restore -o "..$S..$S..$S$(TEMP_DOTNET_DIR)" -f netstandard2.0 ortools$Sdotnet$S$(ORTOOLS_FSHARP_DLL_NAME)$S$(ORTOOLS_FSHARP_DLL_NAME).fsproj
+	$(SED) -i -e "s/MMMM/$(CLR_ORTOOLS_DLL_NAME)/" ortools$Sdotnet$S$(ORTOOLS_NUSPEC_FILE)
+	$(SED) -i -e "s/VVVV/$(OR_TOOLS_VERSION)/" ortools$Sdotnet$S$(ORTOOLS_NUSPEC_FILE)
+	$(NUGET_EXECUTABLE) pack ortools$Sdotnet$S$(ORTOOLS_NUSPEC_FILE)
 
 .PHONY: pkg_dotnet-upload # Upload Nuget Package
 pkg_dotnet-upload: nuget_archive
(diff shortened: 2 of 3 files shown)


Mizux:
thx for this PR.
# Roadmap
## Release
On the longterm I would like to provide a nuget package **with** native libs for all supported platform (like we do for python package). But this is more a requirement for doing a release from my side...
issue: While on python you build a package on each platform then upload them with the same name so package name is "universal" and python package manager retrieve the correct one according to your current host. On the other hand for windows you have to create **one** nuget package which embed the three targets (linux, win, macos).
src: https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks#architecture-specific-folders
## User
For customer we should provide the possibility to create/generate a package from its build directory only containing its platform kind of local custom package build

# Misc
It seems there is also dotnet pack which only use the csproj and not the nuspec to create the package.
I need to dig a little...
Feedback from a dev trying to do it ->
https://stackoverflow.com/questions/49162127/how-to-include-native-assets-in-nuget-on-multiple-platforms-only-64-bit
related issue: https://github.com/NuGet/Home/issues/2372
related issue: https://github.com/NuGet/Home/issues/6645


# Documentation
Here my wish list of RIDs (Runtime IDentifier) target 
`win-x64;linux-x64;osx-x64`
src: https://docs.microsoft.com/en-us/dotnet/core/rid-catalog#using-rids

acco32:
@Mizux For the native libraries, I think it would be cleaner to have a separate installer (with the destination being the system folders). This way, the `dotnet` package is truly platform independent.

mwpowellhtx:
@acco32 @Mizux Would you [consider separating the examples](760) into a separate ``Google.OrTools.Examples.nupkg``? For starters, I am eager to get my hands on at least an RC package that I can start migrating my own assemblies depending on it. Second, this package could carry with it a dependency on the core ``Google.OrTools.nupkg`` itself, and clears the deck to push forward with earnest. Thanks!

jonwagner:
The proposal above seems to match the standard convention for packaging native apps.

* Google.OrTools package
    * Contains .NET library (Google.OrTools.dll)
    * Reference to platform-specific package (e.g. Google.OrTools.linux-x64)
* Google.OrTools.(platform) package (e.g. Google.OrTools.linux-x64)
    * Contains the platform-specific import library (Google.OrTools.Import.dll)
    * Contains platform-specific implementations and dependencies (e.g. Google.OrTools.dylib)

Some notes:

* Don't put anything in the GAC. It's a cause of numerous compatibility issues and is generally unnecessary if you are using nuget packaging.
* I was able to build OSX and Windows versions of the package with the latest code, making small changes to the nupkg file similar to this PR and building them separately. The only issue I saw was including all of the files in the temp_dotnet file brings in a lot of unnecessary files.

mwpowellhtx:
@jonwagner Good point re: GAC. I wasn't aware OR-tools was deploying there? Perhaps I missed something? Anyway, that should be something left to the applications guy(s), per se, minimum integrating OR-tools, if at all. But then again, with the advent of ``netcore``, ``netstandard``, way of doing things, of whole platform/application containers, your run-time, per se, pretty much gets deployed to a well contained sandbox, so it isn't a concern any more.

jonwagner:
OR-tools isn't currently in the GAC, but it was a suggested solution above.

Mizux:
@jonwagner  do you know a public working example of this architecture ?

note: I'm testing on pet project there https://github.com/Mizux/dotnet but still have some difficulties...

jonwagner:
@Mizux here's an article on how to do multi-runtimes as a single package:

https://natemcmaster.com/blog/2016/05/19/nuget3-rid-graph/

The downside is that you get a large package with multiple sets of binaries, but it's really easy to do as a single package.

I'm pretty sure that the large system-level packages split them somehow. Let me poke around and see if I can find an example.

jonwagner:
@Mizux I spent the last half hour looking for documentation and examples for packaging native libraries. They all seem to use the single-package approach.

You build one package with files in a structure like this:

/lib/<framework>/*
/runtimes/<runtimeid>/lib/<framework>/*
/runtimes/<runtimeid>/native/*

The link I posted above has pretty good instructions.

It will make a large package, but then consumers of the package will always have the files it needs. Now that I've looked at it I'd recommend going with the single-package approach.

Mizux:
@jonwagner thanks for the link unfortunately in our case we **must** use the several-packages
 approach which is documented nowhere...
Since or-tools is a C++ library it's nearly impossible to build from source for windows,mac,linux on one platform...

Fortunately after few weeks of effort I manage to do it cf my dev branch...
doc here for details: https://github.com/google/or-tools/blob/mizux/dev/ortools/dotnet/README.md
@all (@mwpowellhtx) I'll try to upload beta package here so you can test it a give me ome feeback if you want...

ToDo before pushing:
- Fix Google.FSharp build (currently disabled)
- Fix Xunit project (currently disabled)
- Move all examples to `examples/csharp` and generate a `.csproj` for each of them.

mwpowellhtx:
@Mizux Blazing the trail. I like it; looking forward to seeing the path you forge.

acco32:
@Mizux the dev branch looks good.

Mizux:
Here a preview if you want to test on Windows **only**:
[Windows-preview-6.8.704.zip](https://github.com/google/or-tools/files/2367204/Windows-preview-6.8.704.zip)

i.e. `Google.OrTools` only depends on `runtime.win-x64.Google.OrTools`

mwpowellhtx:
@Mizux To be clear, I think we got our wires cross. When I think of *Google.OrTools* building for ``netstandard``, I expect the same ``Google.OrTools.$(version).nupkg``. The only difference being the targets. ``Google.OrTools.dll``, perhaps a ``Google.OrTools.xml``, etc, should still be there. I do not see there being any value added introducing yet another package ``runtime.win-x64.Google.OrTools.$(version).nupkg`` whatsoever. Within which ``Google.OrTools.$(version).nupkg`` may contain bitness specifics, as far as I know, but which target is ``netstandard``.

Mizux:
@mwpowellhtx As I said before I can't build one big package for the three OS targeted (win, linux, osx) on one machine so I create a "backend" package for each OS let's call them runtime packages. And one meta package the "frontend" depending on each of them and containing the `Reference assembly`.
- nuget (like maven) package manager contrary to pypi, nix, bintray/conan (<rant>and any smart package mamager</rant>) only want **ONE** archive (i.e. one .nupkg) per version i.e. you can upload one package for each native OS targeted so that's why I had to create a package tree.
- You'll only need to depends on Google.OrTools which transitively depends on runtime packages.
- Since `Google.OrTools` is a meta package it only contains the **reference** assembly in `ref/netstandard2.0/Google.OrTools.dll` you can see it as an empty library only containing a list of "undefined symbol" (if I can make an analogy with the language C) and runtime.* contains the implementation in `runtime/<rid>-x64/Google.OrTools.dll`

Please take a look at https://github.com/google/or-tools/tree/master/ortools/dotnet (feedback/PR welcome)

Mizux:
@acco32 Since Google.OrTools is a meta package, I can't compile any file so FSharp is in a separate package.
Maybe later I will rename the **current** project `Google.OrTools` to `Google.OrTools.CSharp` **then** reuse `Google.OrTools` as empty package depending on `Google.OrTools.FSharp` and `Google.OrTools.CSharp` so customer only need to depends to `Google.OrTools` to get C# and F# which is OK since .Net Core SDK can build both language (contrary to previous mono/.NetFramework which require an extra package for F# (VisualStudio and brew package if I recall well not sure if mono package integrate the F# compiler))

mwpowellhtx:
@Mizux I'm not sure what your sentence *"means, cannot compile any file in it, thus..."* ? As far as I know, to *OrTools* build process, *FSharp* has always been a separate thing over and above and ***depending upon*** *OrTools*.

You've got your dependencies mixed up. Open the packages as ZIP files; you don't even have to rename the extension. You will find *FSharp* depending on the now *empty* package *Google.OrTools*, which now depends on the zero value added *"run time"*. Maybe I am oversimplifying, however, I'd suggest if this is the case that it is time to reconsider how *OrTools* builds.

Given the cross platform nature of *netstandard*, it is easy to include the necessary binaries in the *lib* or even *native* directories, as far as I know. Case in point, let's say someone does want to run *OrTools* in .NET across platforms? Or even embedded? Then what?

Hopefully on the other side of the *Core/Standard* dust settling things will be smoother, but for now, the effort is what it is. But having an entirely separate "runtime" package adds no value IMO but to confuse matters.

I don't know about *Mono*, as I haven't followed it that closely. However, now that things are on the *Core/Standard* track, I don't see anything gained by dwelling there apart from a historical learning perspective.

Thoughts?

Mizux:
I'm closing this pull request since mizux/dev has been integrated to master...

mwpowellhtx:
@Mizux In the context of NuGet packages, right? Come on, for Christ's sake, don't insult my intelligence here. You can open those, extract them, etc, like you would any ZIP file. From there look at the dependencies for yourself in the ***actual*** *nuspec* that got packaged. You can see clearly what the dependencies are.

Re: the Core packages: Ah, I see, good point! :+1: 

Sorry if I came off a bit raw there. This whole *netstandard* thing is a mess and I'm trying to sort some things out by it myself. I do not think my experience therein thus far is too far from being on par with the rest of the industry to date.

Once it is settled and behind us it will be much smoother I'm sure.

Thanks for clarifying.

mwpowellhtx:
One other thing I noticed. Can't wait to see the package [finally published to NuGet](http://www.nuget.org/packages/Google.OrTools#). Approaching that, remember ``Google.Protobuf`` also needs to be potentially repackaged targeting ``netstandard``.

PR #780, opened 2018-08-02
Adding binder links for exploring notebooks in browser

kmader:
A simple badge to let users run the or-tools python package easily in binder
To start jupyter
https://mybinder.org/v2/gh/kmader/or-tools/patch-1
To start a specific notebook
https://mybinder.org/v2/gh/kmader/or-tools/patch-1?filepath=examples%2Fnotebook%2Fgate_scheduling_sat.ipynb


lperron:
Nice, 

Do you know how to update RunFromIPython:

https://github.com/google/or-tools/blob/caca08f3f131f7779335f60705da3b61a4787ae2/ortools/sat/python/visualization.py#L27

to detect we are in binder?

The gate scheduling shoud display a nice graphics.

kmader:
Binder is just jupyter so it shouldn't require any changes from the standard IPython code. Does one of the notebooks have it?

lperron:
actually, it needs plotly and svgwrite. I am debugging the plotly part. 
binder comes with 1.12 by default. I tested with 2.0.5
Latest version is 3.1

And maybe some modules have changed paths.

kmader:
I think the newest version makes it much easier in that you don't need to have the special imports for notebooks and using it offline, with the standard pip install it seems to work

PR #824, opened 2018-08-21
Place all dotnet code in single folder

acco32:
Have all the dotnet code in a single folder


acco32:
@Mizux not all the files are required from that folder (tools/csharp) which is why I didn't remove. And considering how you are planning to change the structure again (more on that later) that may be something you and the other collaborators can assess. There were quite a number of files in the folder regarding the initial netstandard implementation (including a OS specific runtime defintion) which you may want to validate. I've had no issue using the library since the switch to dotnet core. If the feedback from others is positive, I (or others) can start deleting the old folders (ortools/csharp, ortools/csharp, tools/csharp, examples/csharp etc..).

With respect to the change in structure, shouldn't the folders/files be aligned with the language in which they are written? I understand the desire for consistency but should it come at a cost of doing strange things with the language? 

In the meantime, I'll continue to focus on the F# binding. I'll also look at your branch. Thanks for all the help regarding the makefile/build pipelines.

Mizux:
@acco32
- Currently I'm finishing the doc of my Proof of Concept there https://github.com/Mizux/dotnet
(you can already have a nice picture which explain the workflow. then I'll backport/adapt it to for or-tools in my dev branch.
- I think you mean `tools/dotnet` since no files in `tools/csharp` should be in use (here where needed by mono/ Makefile.csharp.mk stuff which has been deprecated/remove)
- Yes on dev branch I try to use only a .csproj since they add few properties to it cf https://docs.microsoft.com/en-us/dotnet/core/tools/csproj
- Still need to learn the purpose of `AssemblyInfo.cs` `CommonAssemblyInfo.cs`	`FlatZincAssemblyInfo.cs` if we can live without it and keep thing simple I would be more than happy.

Also since the CreateSigningKey project is a **tool** replacement of snk.exe (not available/shipped on unix with .NetCore SDK) I'm not sure if we have to put it in ortools/dotnet...
i.e. maybe one day they will add it to .NetCore SDK as global tool...

Mizux:
@acco32 just "recreate" a patch similar on master since this one conflict with my previous change -> I think I will abandon this one...

Issue #856, opened 2018-09-22
Transition to multi-release JAR


kspinka:
Stub issue referenced by macOS fix, for future implementation: http://openjdk.java.net/jeps/238

kspinka:
Newer OS distributions like Ubuntu 18.04 are defaulting to JDK 10 in your internal build system and Visual Studio 2017 to JDK 9.  Please also pin them to Java 8 until a multi-release JAR is ready.

Mizux:
Hi,
- For linux distro, I prefer to stick with the "default-sdk".
- Travis CI vm is a ubuntu 14.04 with so many hack that maybe targeting jdk8 for CI could be a good option
- For Windows I'm thinking about targeting the recommended jdk in the microsoft tutorial https://tutorials.visualstudio.com/Java/hello-world/install-jdk i.e. 8U181 and adapt doc accordingly
- For appveyor CI (windows) need to check which version are provided

BTW you must be able to use `JAVA_HOME` to adapt when you build from source...

kspinka:
The trouble with the default-jdk is that Ubuntu 18.04 will be in a mixed state depending on patch level (pre and post September 2018): https://wiki.ubuntu.com/BionicBeaver/ReleaseNotes

More importantly though, it’s not terribly relevant what version above 8 is utilized on the build image so long as we modify the build to utilize the “--release” flag and/or transition to a multi-release JAR format.

Fundamentally though there are two issues:
1. Since JVMs are (somewhat) backwards compatible with earlier byte code, absent a multi-release JAR, we should prefer to target the least common denominator where no obstruction exists, this is currently Java 8.
2. The Java library build, in general, should not be dependent on the underlying build nor runtime OS.  It will lead to a build matrix explosion and defeat the point of a portable byte code runtime.

Anecdotally speaking, there will be only 2 JVM versions that matter for some time and those are 8 and 11.  So build output targeting only 9 or only 10 just create frustration for users.

Mizux:
@kspinka `should not be dependent on the underlying build nor runtime OS` -> this only work for non native dependent package. OR-Tools is a C++ library wrapped in Java not a pure java library.

Also I need to find a way to provide a native dependent java multi-package in order to be able to generate maven packages. i.e. I want to be able to create a meta-package targeting several native dependent packages like what we did in dotnet cf [here](https://github.com/google/or-tools/blob/master/ortools/dotnet/README.md)

lperron:
Duplicate of #202

PR #857, opened 2018-09-22
Build Java libraries on macOS with Java 8 to improve compatibility.

kspinka:
The default "latest" JDK version coming from Homebrew is a moving target, and is presently using JDK 10.  The world at large is going to be at Java 8 for some time with early adopters leaping to 11.  So instead of grabbing the arbitrary latest version, pin at Java 8 to maximize compatiblity.  This should really be transitioned to a multi-release JAR, see #856, but for now align with the rest of or-tools build policies.


kspinka:
I signed it!

PR #876, opened 2018-10-08
Re-add the CheckedGurobiCall function

### LINKED ISSUE #874: CheckedGurobiCall is not defined
I was trying to use the wrapper for Gurobi mixed integer solver and I get the error -

```
or-tools-6.9.1/ortools/constraint_solver/../gen/ortools/constraint_solver/../../../../lib/libortools.so: undefined symbol: _ZNK19operations_research15GurobiInterface17CheckedGurobiCallEi
```

From the error message it is seemed that the function `CheckedGurobiCall` was not defined.
And looking at the source that function is indeed declared and never defined.

I checked that this bug was introduced in this commit - https://github.com/google/or-tools/commit/4caaef3c23e12841c940345f9e9b91cc367daebf#diff-7f613d0ce96f8e126406a3efb8d1e4f0L168

coderkd10:
fixes #874

### DIFF
--- a/ortools/linear_solver/gurobi_interface.cc
+++ b/ortools/linear_solver/gurobi_interface.cc
@@ -165,7 +165,10 @@ class GurobiInterface : public MPSolverInterface {
   MPSolver::BasisStatus TransformGRBConstraintBasisStatus(
       int gurobi_basis_status, int constraint_index) const;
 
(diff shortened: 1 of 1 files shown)


coderkd10:
I signed it!

PR #910, opened 2018-10-30
Add Constraint return value to methods that were documented to return it

mgritter:
Added "return ct" to three methods.  Ran python tests, but many of them don't use cp_model yet.


Mizux:
note: CI failure was due to a bug with gcc on Travis-CI (patch is in my mizux/master branch)

PR #998, opened 2018-12-30
 Additional guidelines using windows paths added

eliempje:
To avoid make errors due to bad path definitions on windows additional guideline added.
E.g. Adding a \ to an absolute path WINDOWS_SCIP_DIR will result in bad path for SCIP_INC (with space). See tools\make.exe detect_third_party


eliempje:
CLA signed with my username. Everything oke now?

PR #999, opened 2018-12-31
Python path compilations errors fix

eliempje:
When using a python path with spaces (e.g. c:\Program Files\python37) compilations fails. Adding " will solve this issue.


Mizux:
Better to add the guard quote in `PYTHON_INC` directly cf my commit

BTW thanks for the patch ;)

PR #1010, opened 2019-01-08
python: Allow negated literals in LinearExpression

andersk:
Booleans are already considered integers with value 0 or 1, so it’s natural to expect negated booleans to be considered integers with value 1 or 0.

This makes it more convenient to define an antisymmetric array of booleans: you no longer have to choose between not being able to use `greater[j, i] = greater[i, j].Not()` in linear expressions and not being able to use `greater[j, i] = 1 - greater[i, j]` in boolean constraints.


lperron:
Thanks

lperron:
u

Issue #732, opened 2018-06-25
CMake build support future


navidR:
I am trying to add or-tools to vcpkg [1].  But before that, I need to know what is future of cmake support in or-tools? I see it is experimental. Experimental means there is a good chance of removing it in future? Or experimental in this context means it is going to get more stable and eventually fully supported in or-tools? (In this case, I can rely on CMake support of or-tools team and add it to vcpkg).

1) https://github.com/Microsoft/vcpkg/issues/692

lperron:
We would like to drop make in favor or cmake completely.

There are still a few issues, but this is the future of the build part of
or-tools.

Thanks
Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00



Le lun. 25 juin 2018 à 07:30, Navid Rahimi <notifications@github.com> a
écrit :

Mizux:
Hi,

Yes it is intended to use CMake as default build (at least for C++) and to replace our current Makefile based system.
For vcpkg it's a long term goal but before we ant or-tools deps to be also in vcpkg IMHO.

i.e. currently in CMakelists.txt I build zlib, gflags, glog, protobuf, mizux/CBC and merge everything.
I would like to depends on vcpkg instead but:

ZLIB: want to use my CMake version https://github.com/madler/zlib/pull/337
GFlag: better if we could use gflags::gflags https://github.com/gflags/gflags/pull/247
CBC: 
- on github it is split in 5 repos (CoinUtils, Osi, CLP, CGL, CBC) we now use this layout for unix since most distro use this instead of the svn standalone CBC build
- there i no support for CMake so I did one here https://github.com/Mizux/Cbc but i use the svn one repo version instead of the splited ones -> need to redo it

Then we will provide an OR-Tools vcpkg after if you want to work on our dependencies to vcpkg them all...

CMake is currently experimental since:
- dependencies management is not still clear (currently we `add_subdirectory()` all our deps using `AddProject` tricks)
- Currently I only did C++ / python swig support. I need to figure out how to do the .Net and Java support since CMake is a C/C++ project generator at its core.
e.g. CMake learn C# in CMake 3.8 https://cmake.org/cmake/help/v3.8/release/3.8.html ...
- We would like to here from community (tester) if my CMake support is stable enough and work for them

navidR:
It is good news to hear we are going to rely on CMake. And only C++ would suffice I think.

[...]

PR #1019, opened 2019-01-18
Feature/better find package

victor-uavia:
This PR is a proposal to fix https://github.com/google/or-tools/issues/732 using `find_package` in order to have better granularity for handling dependencies (this is inspired by https://gist.github.com/Som1Lse/6458d28c4bb09163a5389f24fec981db).
Thanks to the built-in variable `<package_name>_DIR` variable we can override the default path for `find_package` in order to use : 
- some preinstall cmake package
- custom source repo (either download with `BUILD_DEPS=ON` or already download and provide by the user through a custom `<Package_name>Config.cmake`)

Allowing to use `find_package` in order to resolve out of tree dependencies allow or-tools to integrate nicely with Yocto, which is not the case today as Yocto since we cannot relay on the `BUILD_DEPS` switch in Yocto (it does not allow to download dependency during cmake configure time).

⚠️ Note that this PR will properly for work preinstall cmake package once https://github.com/abseil/abseil-cpp/pull/182 will be merged


victor-uavia:
@Mizux do you think this PR could be integrate into the code ?

Mizux:
First, thanks for your contribution !
 
If I understand correctly, you *just* split the `cmake/external/CMakeLists.txt` into several `DependencyConfig.cmake` so for each deps you could choose the system package or ortools will compile it for you.
-> I'm ok with this.

*But* in this case the new macro `_find_package()` seems strange to me 
I mean you `include()` them if needed, while for me you should have modified the `CMAKE_PREFIX_PATH` instead so `find_package()` would find them...

Also, one other side topic: I think I'll have to change the CMake behavior, I would like to install deps first then consume the install, than trying to use it as `add_subdirectory()`. Currently it's conflict with dependency install rule or you need to hack dependencies to make it work...
see: https://travis-ci.org/google/or-tools/jobs/482350303#L904

Sorry for the delay, I'm still thinking about it and where I would like to go.
Need to play with your branch...

CMakeLists.txt
```diff
@@ -79,25 +70,25 @@ message(STATUS "Build .Net Binding: ${BUILD_DOTNET}")
...
-  set(ZLIB_DIR ${CMAKE_CURRENT_SOURCE_DIR}/cmake/external CACHE PATH "ZLIB dependency path")
+  list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/external/ZLIB)
 endif()
 if(NOT Protobuf_DIR AND BUILD_DEPS)
-  set(Protobuf_DIR ${CMAKE_CURRENT_SOURCE_DIR}/cmake/external CACHE PATH "Protobuf dependency path")
+  list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/external/Protobuf)
 endif()
 if(NOT abseil_DIR AND BUILD_DEPS)
-  set(abseil_DIR ${CMAKE_CURRENT_SOURCE_DIR}/cmake/external CACHE PATH "abseil dependency path")
+  list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/external/abseil)
```
Mizux: abseil-cpp ? (I can also do it in a later commit) 
just becaue there is also https://github.com/abseil/abseil-py and maybe we will need it one day too

victor-uavia:
@Mizux I rework the way this branch handle its dependencies in https://github.com/google/or-tools/pull/1019/commits/2a003e0938c15c885ee9976835d22568bbc8534a the branch now donwload, build & *install its file* in the build directory (no more subproject).

I have however the following compilation error which I do not understand and I would like your help to see what I could have missed.

```
/or-tools/or-tools/ortools/data/jobshop_scheduling_parser.cc:26:1: error: C++ requires a type specifier for all declarations
DEFINE_int64(jssp_scaling_up_factor, 100000L,
^
/or-tools/or-tools/ortools/data/jobshop_scheduling_parser.cc:26:14: error: use of undeclared identifier 'jssp_scaling_up_factor'
DEFINE_int64(jssp_scaling_up_factor, 100000L,
             ^
/or-tools/or-tools/ortools/data/jobshop_scheduling_parser.cc:359:45: error: use of undeclared identifier 'FLAGS_jssp_scaling_up_factor'
          static_cast<int64>(round(weight * FLAGS_jssp_scaling_up_factor));
                                            ^
/or-tools/or-tools/ortools/data/jobshop_scheduling_parser.cc:375:19: error: use of undeclared identifier 'FLAGS_jssp_scaling_up_factor'
                  FLAGS_jssp_scaling_up_factor !=
                  ^
/or-tools/or-tools/ortools/data/jobshop_scheduling_parser.cc:385:17: error: use of undeclared identifier 'FLAGS_jssp_scaling_up_factor'
                FLAGS_jssp_scaling_up_factor);
                ^
/or-tools/or-tools/ortools/data/jobshop_scheduling_parser.cc:389:22: error: use of undeclared identifier 'FLAGS_jssp_scaling_up_factor'
              1.0L / FLAGS_jssp_scaling_up_factor);
                     ^
```

Mizux:
Hi,

I revamped your CL in #1116 but with few changes:
* Mostly Keep `BUILD_dependencies` options BUT
  * by default don't build dependencies (easier for distro package maintainer) **unless** you build wrappers
   (ed BTW [PEP 513](https://www.python.org/dev/peps/pep-0513/#the-manylinux1-policy) for python native package doesn't allow us to use system libraries, so if we have to ship them...)
  * you can still enable `BUILD_*` if you want...
  * if BUILD_DEPS=ON all `BUILD_*` options are forced to ON
    https://github.com/google/or-tools/blob/4477092d8d38854b19b577bd586fd2829d0e8838/CMakeLists.txt#L49-L50
   which is the opposite of your
   https://github.com/victor-uavia/or-tools/blob/2a003e0938c15c885ee9976835d22568bbc8534a/CMakeLists.txt#L42-L43

* I don't use a `Find*.cmake` simply call `find_package(* REQUIRED)` and let CMake to [do its stuff](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure)
  * user can use `CMAKE_PREFIX_PATH`/`CMAKE_MODULE_PATH` if they want to hack etc..

* all deps, if build, are installed in `CMAKE_BINARY_DIR/dependencies/install/` so I just need to append this to `CMAKE_PREFIX_PATH` to see all available prebuilt dependencies.

* I use a template https://github.com/google/or-tools/blob/master/cmake/dependencies/CMakeLists.txt.in for ExternalProject along a function in utils.cmake
  https://github.com/google/or-tools/blob/4477092d8d38854b19b577bd586fd2829d0e8838/cmake/utils.cmake#L66-L126

* I use my new Cbc split repo forks of the official ones

* I use the fixed abseil-cpp name i.e. `absl::base` instead of "broken" `absl::absl_base` since my internal path make its way to the public abseil-cpp repo.

Any feedback are welcome, hope you'll better understand what I'm aiming for....

Mizux:
I'm closing this PR since we have #1116 
Still open to any proposal, critics etc....

PR #1072, opened 2019-02-19
np.random.randn to np.random.randint

tk120404:
ref: https://docs.scipy.org/doc/numpy-1.16.1/reference/routines.random.html


Mizux:
1) Why do you want to change from a "standard normal" distribution to a "uniform" one ?

2) Can you please do your PR against master branch instead ?

tk120404:
Sure will do. Closing this

PR #1119, opened 2019-03-11
Remove references to tsp and vrp projects from examples solution file

stuarthillary:
Those projects are in a different directory


Mizux:
Thx

PR #1120, opened 2019-03-11
Port the C++ example network_routing_sat to C#

stuarthillary:
Port of [network_routing_sat.cc](https://github.com/google/or-tools/blob/master/examples/cpp/network_routing_sat.cc) to .Net Standard 2.0


Mizux:
Waiting for CI but LGTM.
Thx again ! :wink:

stuarthillary:
Running the model with the default parameters gives an assertion failure. What is the best approach, adjust the default values in code or pass in some sensible values?

Mizux:
Usually we try to keep sync between languages on data model, default values etc...
in C++ we have:
https://github.com/google/or-tools/blob/2b36515b43e6ce11d26e3a8ac8e97f30d66a3ccf/makefiles/Makefile.cpp.mk#L469-L471

so I would suggest to use the same `ARGS` value.

Mizux:
dotnet pass on Travis-CI (see https://travis-ci.org/google/or-tools/jobs/505228561) -> I'll merge

Issue #1169, opened 2019-03-29
cbc.CMakeLists.txt should point to a version in or-tools releases


moriarty:
in v6.10:

https://github.com/google/or-tools/blob/554cbccaa95b4c11eced13f145de1bf468e1f919/cmake/external/cbc.CMakeLists.txt#L12

the tag points to master

moriarty:
This is almost a duplicate of #1167 and #1159 ...

The fix for #1159 https://github.com/google/or-tools/commit/e08d64c370bb02af07b463ee951be74353e79e51 is still brittle because it points to the master branch, and anyone building a tagged version of or-tools who would expect the build to be the same won't always get the same result if changes are pushed to cbc master.

If or-tools is being build in a docker layer, for example:

```Dockerfile
ARG OR_TOOLS_RELEASE_TAG=v6.10

RUN git clone --branch ${OR_TOOLS_RELEASE_TAG} --single-branch --depth=1 \
      https://github.com/google/or-tools /var/local/git/or-tools && \
    cd /var/local/git/or-tools && \
    mkdir build && \
    cd build && \
    cmake .. && \
    make -j$(nproc) && \
    make install && \
    cd / && \
    rm -rf /var/local/git/or-tools
```

Then this is now broken.

moriarty:
work around for now (diff of above Dockerfile):

```diff
 ARG OR_TOOLS_RELEASE_TAG=v6.10
 
+# We need a patch to fix an issue in an OR-Tools dependency:
+# https://github.com/google/or-tools/issues/1169
+
 RUN git clone --branch ${OR_TOOLS_RELEASE_TAG} --single-branch --depth=1 \
       https://github.com/google/or-tools /var/local/git/or-tools && \
     cd /var/local/git/or-tools && \
+    curl https://github.com/google/or-tools/commit/e08d64c370bb02af07b463ee951be74353e79e51.patch | git apply -v && \
     mkdir build && \
     cd build && \
     cmake .. && \
```

Mizux:
the issue is not master branch vs commit sha1, it is because I have to reuse the repository name Cbc so I rename it coinor-cbc (which I will archived if we can still clone it)...

Previous way to build Cbc was based on the coinor svn archive where I added a full CMake support but it was based on this monolithic archive while you can see now on github (and in few distro) that coin or change their mine and provide it as separate repo (CoinUtils, Clp, Cbc ...).

[...]

PR #1170, opened 2019-03-29
Lock cbc to specific commit and avoid flaky builds

lnikkila:
Fixes https://github.com/google/or-tools/issues/1169


Mizux:
https://github.com/Mizux/coinor-cbc is archived but if you want a SHA1, i'm OK with this and validate this short term PR.

PR #1222, opened 2019-04-26
Patch RECORD file after patch WHEEL.

georgi-yanchev:
Updates size and hash of WHEEL file after it was patched.

@Mizux please advice whether this should be merged into `stable` or some other branch.

### DIFF
--- a/tools/docker/build-manylinux1.sh
+++ b/tools/docker/build-manylinux1.sh
@@ -187,8 +187,17 @@ do
     # supported by pip
     FILE=(${EXPORT_ROOT}/*-${PYTAG}-*.whl)
     unzip "$FILE" -d /tmp
-    sed -i 's/manylinux2010/manylinux1/' /tmp/ortools-*.dist-info/WHEEL
-    rm -f $FILE
+    FILE_TO_PATCH=$(ls /tmp/ortools-*.dist-info/WHEEL)
+    RECORD_FILE=$(ls /tmp/ortools-*.dist-info/RECORD)
+    # Save old hash and size, in order to look them up in RECORD
+    OLD_HASH=$(sha256sum ${FILE_TO_PATCH} | cut -c -64 | xxd -r -p | base64 | tr +/ -_ | cut -c -43)
+    OLD_SIZE=$(wc -c < ${FILE_TO_PATCH})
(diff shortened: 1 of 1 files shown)


georgi-yanchev:
I signed it!

Mizux:
Hi,

Sorry for the delay.

I would like the thnak you for this PR even if I will reject this PR since I have done 3f0a13ca47bfc6d41e0ddf5c234d4fb0de9a2a49 (mostly based on your work)
Few points:
* branch **master** is our devel branch, so please make a PR against it (next time ?), default branch **stable** is our release branch which bump from one release tag to an other...
* In your PR you use:
    ```sh
    $(sha256sum ${FILE_TO_PATCH} | cut -c -64 | xxd -r -p | base64 | tr +/ -_ | cut -c -43)
    ```
    I feel it, a little bit ad hoc without any reference to the python code (e.g. why using `tr` or `-43`) -> I've  prefered to try to use the same code used by setuptools.py -> rewrite it using a small python script
* also the first `cut -c -64` should have be `awk -F '{print $1}'` IMHO
* In your PR there is no test/check that the sha256sum fix is correctly working -> I added a quick test using `pex` directly (first, I manage to reproduce your error by removing the *sha256=* sed)

BTW thanks a lots for this PR, don't hesitate to get back to me if you have some comments etc...

note: there was also a bug on master on Centos7, glog libraries which is installed in `lib64` dir was not copied IIRC

georgi-yanchev:
Thanks. I'm glad that this got resolved.

georgi-yanchev:
In case you are still wondering:

* `tr +/ -_` because WHL files use URL-safe base64: https://en.wikipedia.org/wiki/Base64#Variants_summary_table

* 43 is the number of base64 characters needed to encode 256 bits with trimmed `=` padding: ceil(256/8 * 4/3)

Mizux:
@georgi-yanchev yes I know, I tried to reverse the setuptools.py on last Monday to figure out too ;)

https://github.com/pypa/wheel/blob/64e089fd175e38e0ea77975131384960c6f2223d/wheel/util.py#L26-L28
and
https://github.com/pypa/auditwheel/blob/d99cc1dea74207dc6891085bbb65c87538014d4f/auditwheel/wheeltools.py#L85-L88

Issue #202, opened 2016-06-29
Add Maven Support 


danyaljj:
Original post
I can see some maven releases here: 
https://bintray.com/nikola-ilo/maven/or-tools/view
Q: Why the releases are not continued? Are you guys planning to release a new version on maven? 
A: Not done by Google LLC

For v8.0 we are migrating to use maven to create java package
Todo List:
* [x] Use Maven in CMake based build
  * [x] Add Maven package to CMake based build
  * [x] Use Maven for samples (`ortools/*/samples`)
  * [x] Use Maven for examples  (`examples/contrib examples/java`)
  * [x] Use Maven for tests (`examples/tests`)
* [x] Use Maven in Make based build
  * [x] Add Maven packages to Make based build
  * [x] Use Maven for samples
  * [x] Use Maven for examples
  * [x] Use Maven for tests
* [x] Rework archive to use Maven packages
  * [x] Linux archive
  * [x] MacOS archive
  * [x] Windows archive
* [x] Add internal doc in:
  * [x] ortools/java/README.md
  * [x] ortools/java/doc/local_pipeline.dot
  * [x] cmake/doc/java.md

lperron:
These are actual contributed and we did not do it.
I will contact the author to see if we can transfer it to or-tools.

Thanks

Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00


2016-06-30 0:20 GMT+02:00 Daniel Khashabi <notifications@github.com>:

PunKeel:
I would definitely appreciate being able to integrate or-tools using Maven

dlahlou:
Hello Punkeel,
It's on our to-do list to support Maven. Thank you.
Regards,
Driss

On Aug 26, 2016 03:36, "PunKeel" <notifications@github.com> wrote:

I would definitely appreciate being able to integrate or-tools using Maven

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/google/or-tools/issues/202#issuecomment-242597370>, or mute
the thread

guizmaii:
Any news ?

fcorneli:
Maven support would be interesting...

Mizux:
hi all,

Yes I would like to provide a maven (gradle ? ivy ?) package for or-tools ASAP and ideas are welcome !

How I see it currently:
1.Have a clean and modern CMake project to replace Makefile (cf WIP mizux/cmake branch)
[...]

PR #1265, opened 2019-05-10
Java: also build a jar file containing sources

duchampdev:
Having a sources jar noticeably eases development of or-tools based applications, especially if done trough an IDE.

### DIFF
--- a/makefiles/Makefile.java.mk
+++ b/makefiles/Makefile.java.mk
@@ -300,6 +300,7 @@ $(JAVA_OR_TOOLS_LIBS): \
  $(GEN_PATH)$Scom$Sgoogle$Sortools$Sgraph$S*.java \
  $(GEN_PATH)$Scom$Sgoogle$Sortools$Slinearsolver$S*.java
 	"$(JAR_BIN)" cvf $(LIB_DIR)$Scom.google.ortools.jar -C $(CLASS_DIR) com$Sgoogle$Sortools$S
+	"$(JAR_BIN)" cvf $(LIB_DIR)$Scom.google.ortools-sources.jar -C $(GEN_DIR) com
 
 ###################
 ##  Java SOURCE  ##


Mizux:
not against the idea but if its small enough why not directly add them to the original or-tool.jar ?

duchampdev:
Thanks for your fast reply!
I just checked the size, it's just about 430 KB. So in fact your suggestion seems quite sensible.
My initial intention was to keep the changes to existing code/behaviour as little possible. Furthermore, I'm quite used to keep the sources separate from Maven.
But yes, I'm pleased to add another commit to combine everything in a single jar.

duchampdev:
Update: While looking for the best way/place to include the sources, I stumbled upon this: https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javac.html#searching.
In summary, including sources into the binaries jar may in some cases lead to double compiled files when including the or-tools lib in other projects which hence can interfere with or crash the classloader.
I guess there are valid arguments for both ways, especially w.r.t. the linked article. Which way would you prefer @Mizux?

makefiles/Makefile.java.mk
```diff
@@ -300,6 +300,7 @@ $(JAVA_OR_TOOLS_LIBS): \
  $(GEN_PATH)$Scom$Sgoogle$Sortools$Sgraph$S*.java \
  $(GEN_PATH)$Scom$Sgoogle$Sortools$Slinearsolver$S*.java
 	"$(JAR_BIN)" cvf $(LIB_DIR)$Scom.google.ortools.jar -C $(CLASS_DIR) com$Sgoogle$Sortools$S
+	"$(JAR_BIN)" cvf $(LIB_DIR)$Scom.google.ortools-sources.jar -C $(GEN_DIR) com
```
lperron: Not sure this contains all the sources. I guess you need to include generated sources, they will be on genS$com

duchampdev: Hi @lperron , sorry for my very late reply! I was very busy last year from summer on :(.
The command actually executed is: `"/usr/lib/jvm/java-1.8.0-openjdk/bin/jar" cvf lib/com.google.ortools-sources.jar -C ortools/gen com`. So if I understand your comment correctly, this should include the generated sources. Is this correct or do I miss something?

oliviercailloux:
Perhaps you could consider raising priority of making a Maven release? This would solve this issue as well (hence save total time). And it would make it easier to follow good practices. And I suspect it would be not too hard. And, really, it’s quite unusual for a good quality Java open source project

Mizux:
@oliviercailloux Actually is my top priority once the release v7.6 is out, I should have postponed it to v7.7 my bad.

I'm currently having trouble to fine good documentation on integrate native library in pom.xml and having a good loader.
So #202 is definitively not dead !
ps: I'll first try on pet project mizux/cmake-swig project which is way faster to build (no third party but still the whole swig stuff)

lperron:
It is very hard to do a multi-platform maven release with native libraries
:-(
00



Le mer. 22 avr. 2020 à 16:50, Olivier Cailloux <notifications@github.com> a
écrit :

oliviercailloux:
@Mizux great news!

See https://github.com/oliviercailloux/google-or-tools-java/issues/2 for my position about native libraries. Very unfortunate difficulty in Java indeed.

Mizux:
obsolete since we are using maven to generate java packages see #202

duchampdev:
@Mizux Thanks for the information. Really happy to see this PR getting obsoleted in this way (Maven).

PR #1352, opened 2019-06-12
Incorporate all public methods of KnapsackSolver into Java wrapper

piyo7:
I want to use full features of this great knapsack solver from JVM, therefore I modify the Java SWIG interface referring to #1097.
Additional methods are renamed to lower camel case.


lperron:
Thanks

Issue #1399, opened 2019-07-03
Possible bugs on shift_scheduling_sat.py example


stradivari96:
I think
```
for start in range(len(works) - length - 1):
```
should be
```
for start in range(len(works) - length + 1):
```

> https://github.com/google/or-tools/blob/master/examples/python/shift_scheduling_sat.py#L95

For a list of length 4 the possible sequences of length 2 are:
```
[1, 1, 0, 0], [0, 1, 1, 0], [0, 0, 1, 1]
```

The last start is 2 -> range(4-2+1) -> range(3) -> [0, 1, 2]

There are many lines where this appears, so maybe I am wrong.

Could test it and send a pull request.

lperron:
It is a start, if the length is 2, the last element is 2 + 2 - 1, which is
the index of the last element in a zero based array of size 4.

Am I wrong?
Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00



Le mer. 3 juil. 2019 à 23:58, Xiang Chen <notifications@github.com> a
écrit :

stradivari96:
Didn't understand the last message very well, but right now given a list of length 4 and a sequence length of 2, the line:
```
for start in range(len(works) - length - 1):
```
will only iterate through `range(4-2-1)` which is `range(1) == [0]` when I think it should return `range(3) == [0, 1, 2]`, not sure where the first 2 in `2 + 2 - 1` comes from.

lperron:
Actually, both are wrong

you have 4 variables

v0, v1, v2, v3

You want to forbid sequence of exactly 2 true variables, you need to
construct 3 forbidden patterns

v0=true, v1=true, v2=false
v0=false, v1=true, v2=true, v3=false
v1=false, v2=true, v3=true

The negated bounded span takes care of adding the var=false to the pattern,
so the valid starts are 0, 1, 2. Therefore, both +1 and -1 are wrong.

Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00



Le jeu. 4 juil. 2019 à 10:58, Xiang Chen <notifications@github.com> a
écrit :

lperron:
Thanks for the report. I have fixed the example on master.

stradivari96:
I still think it should be +1 though.
> You want to forbid sequence of exactly 2 true variables, you need to construct 3 forbidden patterns

[...]

PR #1403, opened 2019-07-04
Fixes to ShiftSchedulingSat.cs example in the same way as Python fixes

stuarthillary:
Fixes to C# example in the same way as fixes in changeset 4a0e9b1860276a021335aacb8b69e10a0d08942c for issue https://github.com/google/or-tools/issues/1399


stradivari96:
I would also remove the -1 in:
https://github.com/google/or-tools/blob/b8b7a5b196676966ef66351c4f83571f9b285a85/examples/dotnet/ShiftSchedulingSat.cs#L457

https://github.com/google/or-tools/issues/1399#issuecomment-508420736

lperron:
Thanks

PR #1424, opened 2019-07-16
fix: update cmake protobuf version

matthewrj:
This aligns the version of protobuf used by cmake to the same as what is defined in Dependencies.txt. Currently using BUILD_DEPS=ON with cmake builds the wrong version of protobuf (3.7.1) and so when the resulting ortools binary is used, protobuf throws an error due to a mismatch between the version of protoc used to generate the protobuf source files and the runtime version.


matthewrj:
I signed it!

lperron:
Thanks

PR #1431, opened 2019-07-18
Expose swap index pair as its own local search operator

matthewrj:
This PR exposes "swap index pair" as its own independent local search operator. It currently exists as part of "node pair swap".

I have experienced significant speedups by using "swap index pair" by itself and not "node pair swap". "node pair swap" seems to be quite computationally expensive operator and "swap index pair" much cheaper while still being powerful as a search operator.

The use case this has been tested under is pickup-delivery with time windows and capacity where each pickup and delivery have multiple options (disjunctions).


lperron:
Small comment. We have not forgotten this. It is just a bit complex to make sure it fits in our global framework.

matthewrj:
Ok thanks for the update

matthewrj:
Can you please close this PR if it is never going to be merged.

PR #1493, opened 2019-08-13
Change the order of variable declaration to save memory

mingodad:
Changing the order of variable declaration we can waste less memory on memory alignment generated by the compiler.
In one test problem with 9,320,328 variables and 1,107 constraints it saves 10% of memory usage.


stradivari96:
Just wanted to point out that is was opened against stable instead of master, not sure if that is a problem or not

lperron:
Ah s**t, can you reopen against master ?
00



Le mar. 13 août 2019 à 16:12, Xiang Chen <notifications@github.com> a
écrit :

PR #1541, opened 2019-09-03
Xpress interface

bzfberth:
This includes a few fixes to the use of Xpress controls (which in some cases differ to the usage in other solvers).

Namely, time limits should usually be given as negative numbers (positive numbers will only cause a stop when a solution has already been found,
scaling is a bit field control, not an on/off switch, and
outputlog is the control to use for deactivating all output. Lplog and miplog are merely there to change frequency and style of output.


lperron:
Integrated manually.
Closing the PR.

Thanks

PR #1544, opened 2019-09-03
Add .pyi files for CP-SAT to enable autocompletion

stradivari96:
So I generated some stub files using https://github.com/dropbox/mypy-protobuf.

This allows some IDEs such as PyCharm to autocomplete and infere types:

Screenshot from 2019-09-03 22-38-33
Screenshot from 2019-09-03 22-38-52

Want this added to the protoc flow?


stradivari96:
Actually, the right thing to do instead of adding these files would be:
* pip install mypy-protobuf
* use protoc with --mypy_out= option

lperron:
Thanks, I will have a look.
00



Le mar. 3 sept. 2019 à 22:54, Xiang Chen <notifications@github.com> a
écrit :

Mizux:
Should be doable in CMake
> Actually, the right thing to do instead of adding these files would be:
> 
> * pip install mypy-protobuf
should be done like this:
https://github.com/google/or-tools/blob/84c35244c9bdd635609af703bbf3053c16c487c0/cmake/python.cmake#L133
replacing `wheel` by `mypy-protobuf`
> * use protoc with --mypy_out= option
add it there:
https://github.com/google/or-tools/blob/84c35244c9bdd635609af703bbf3053c16c487c0/cmake/python.cmake#L50-L54

PR #1555, opened 2019-09-07
Feature/python sports schedule example

jmarca:
Translated sports_schedule_sat.cc to python. 

Added command line options to vary the numbers of teams, number of game days, etc., plus the ability to dump the resulting schedule to a CSV file.

Ran the code through  `yapf --style "google"`

No tests.  

Will not work with an odd number of teams, as "bye" days are not computed.


jmarca:
Oops.  Not right.

jmarca:
Sorry, it seems I made a PR with too many commits!

PR #1556, opened 2019-09-07
Translate sports_scheduling_sat.cc to python

jmarca:
More or less a straight translation of the C++ to python.  Originally
developed in response to a question on the OR Tools forums, so there
are more options than the C++.  

Note there are no tests of this code.  Also it will not work for odd number of
teams.


lperron:
Thanks

PR #1557, opened 2019-09-07
Update hinting examples

stradivari96:
Nothing much.


lperron:
Thanks. Now, I need to write this examples for the other languages :-)

stradivari96:
I think it also will be useful make this example a little bit more complex, as it gets directly to the optimal without any hinting right now

lperron:
This is really not the point of the example. Just the syntax.
00



Le sam. 7 sept. 2019 à 17:35, Xiang Chen <notifications@github.com> a
écrit :

Issue #1547, opened 2019-09-04
Use mypy-protobuf to enable type hinting and autocompletion


stradivari96:
https://github.com/google/or-tools/pull/1544 makes more sense as an feature request.
EDIT: #1544 has been dropped

This allows some IDEs such as PyCharm to autocomplete and infere types:

![Screenshot from 2019-09-03 22-38-33](https://user-images.githubusercontent.com/19752586/64207252-8c905500-ce9c-11e9-8839-cf1a754554e0.png)
![Screenshot from 2019-09-03 22-38-52](https://user-images.githubusercontent.com/19752586/64207261-9023dc00-ce9c-11e9-9587-811cdfa8854a.png)

To enable this:
* [x] CMake
  * [x] Add pip install mypy-protobuf
  * [x] Use protoc with --mypy_out= option
  * [x] Add mypy-protobuf in ci docker
* [x] Make
  * [x] Add pip install mypy-protobuf
  * [x] Use protoc with --mypy_out= option
  * [x] Add mypy-protobuf in ci docker
* [x] Add mypy-protobuf in Travis CI
* [x] Add mypy-protobuf in Appveyor CI
* [x] Add mypy-protobuf in tools/docker
* [x] Package `*.pyi` files (see: https://github.com/google/or-tools/pull/1640/commits/2515e62c87af7533d56ec4a081c5cf90500d0277)

Source:
https://github.com/dropbox/mypy-protobuf

Mizux:
The *.pyi files are used by PyCharm and other development tools to provide
more information, such as PEP 484 type hints, than it is able to glean from
introspection of extension types and methods.  They are not intended to be
imported, executed or used for any other purpose other than providing info
to the tools. If you don't use use a tool that makes use of .pyi files then
you can safely ignore this file.

See: https://www.python.org/dev/peps/pep-0484/
     https://www.jetbrains.com/help/pycharm/2016.1/type-hinting-in-pycharm.html

Mizux:
I reopen it since #1544 have been dropped...

Mizux:
Last comment on it
https://github.com/google/or-tools/pull/1544#issuecomment-721703458

Mizux:
now should be fixed, still need to verify release script are working as expected I.e. double check we generating the incoming v8.1

PR #1640, opened 2019-10-12
Distribute mypy files

stradivari96:
Fix https://github.com/google/or-tools/issues/1547

The *.pyi files were created (https://github.com/google/or-tools/commit/e1b32efb5af2fc6b486ca98528ef3f985f3e4b44) but not added to the wheel, this PR modifies `setup.py` to include them.

### DIFF
--- a/makefiles/Makefile.python.mk
+++ b/makefiles/Makefile.python.mk
@@ -843,36 +843,31 @@ clean_python:
 	-$(DEL) $(GEN_PATH)$Sortools$S__init__.py
 	-$(DEL) ortools$S*.pyc
 	-$(DELREC) ortools$S__pycache__
-	-$(DEL) $(GEN_PATH)$Sortools$Salgorithms$S*.py
-	-$(DEL) $(GEN_PATH)$Sortools$Salgorithms$S*.pyc
+	-$(DEL) $(GEN_PATH)$Sortools$Salgorithms$S*.py*
 	-$(DELREC) $(GEN_PATH)$Sortools$Salgorithms$S__pycache__
 	-$(DEL) ortools$Salgorithms$S*.pyc
 	-$(DELREC) ortools$Salgorithms$S__pycache__
 	-$(DEL) $(GEN_PATH)$Sortools$Salgorithms$S*_python_wrap.*
 	-$(DEL) $(GEN_PATH)$Sortools$Salgorithms$S_pywrap*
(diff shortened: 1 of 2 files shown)


stradivari96:
Gentle ping.

PS: more context https://stackoverflow.com/questions/34571088/how-to-distribute-type-hints-to-pypi

lperron:
I did not manage to make it work on all platforms.

stradivari96:
Are you talking about the `protoc-gen-mypy` part? On which platforms?
I can take a look but a simple fix could be to just commit the generated .pyi files

Mizux:
Does the ortools .pyi files shouldn't be part of the [typeshed](https://github.com/python/typeshed) repo instead (like [protobuf](https://github.com/python/typeshed/tree/master/third_party/2and3/google/protobuf)) ?

stradivari96:
Didn't know about that repo, but I believe or-tools may be a bit too specific to be included there.
Your protos also tend to change each version with new parameters so keeping that in sync could be a bit hard.

stradivari96:
Investigating a bit more, it might be easier to just copy/clone [the scripts](https://github.com/dropbox/mypy-protobuf/tree/master/python) and use the other method indicated on the [README](https://github.com/dropbox/mypy-protobuf#python-implementation).

> Alternately, you can explicitly provide the path:
> ```
> protoc --plugin=protoc-gen-mypy=path/to/protoc-gen-mypy --python_out=output/location --mypy_out=output/location
> ```
> On windows, provide the bat file:
> ```
> protoc --plugin=protoc-gen-mypy=path/to/protoc_gen_mypy.bat --python_out=output/location --mypy_out=output/location
> ```

That should make it work across platforms and you can avoid the `pip install` (the only requirement is `protobuf>=3.6.0`)

stradivari96:
Closing this for now, will reopen when I can work on it

stradivari96:
Hi! I created and uploaded a package called [ortools-stubs](https://pypi.org/project/ortools-stubs/) to pypi similar to https://github.com/numpy/numpy-stubs, following [PEP-561](https://www.python.org/dev/peps/pep-0561/).

This way, running `pip install ortools-stubs` downloads all the `*.pyi` files enabling autocompletion (at least in Pycharm) for every platform.

Links:
* https://pypi.org/project/ortools-stubs/
* https://github.com/stradivari96/ortools-stubs

Hit me up if you want me to transfer the ownership of the pypi package.

PR #1728, opened 2019-11-20
Patch 15

paultrow:
Adding MipVarArray doc examples.


Mizux:
The only relevant commit is the add of `MIPVarArray.cs` which is already on master so I think it makes its way through an other path -> safely close this merge

PR #1835, opened 2020-01-09
Update ISSUE_TEMPLATE

Mizux:
Use link instead of green button for Modeling issues...


Mizux:
@googlebot I signed it!

PR #1850, opened 2020-01-20
Added java convenience method to get domain of IntVar

raphweb:
Sorry for the confusion, previous pull request had wrong author email address.

### DIFF
--- a/ortools/java/com/google/ortools/sat/IntVar.java
+++ b/ortools/java/com/google/ortools/sat/IntVar.java
@@ -19,11 +19,12 @@ import com.google.ortools.util.Domain;
 
 /** An integer variable. */
 public final class IntVar implements Literal, LinearExpr {
-  IntVar(CpModelProto.Builder builder, Domain domain, String name) {
+  IntVar(final CpModelProto.Builder builder, final Domain domain, final String name) {
     this.modelBuilder = builder;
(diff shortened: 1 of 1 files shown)


ortools/java/com/google/ortools/sat/IntVar.java
```diff
@@ -19,11 +19,12 @@
 
 /** An integer variable. */
 public final class IntVar implements Literal, LinearExpr {
-  IntVar(CpModelProto.Builder builder, Domain domain, String name) {
+  IntVar(final CpModelProto.Builder builder, final Domain domain, final String name) {
```
Mizux: since java 8, final is implicit for arguments so you don't need to write it explicitly.

raphweb: True, but there is a difference between effectively final and final. I usually use the latter if possible.
Anyway, you can keep the old version.

lperron:
I will close this PR. I am in the process of re-implementing. The problem is that you can manipulate the inner proto and thus have the stored domain and the proto diverges. I will implement a version that rebuilds  the domain from the proto.

raphweb:
Thanks a lot :-)

lperron:
I just pushed the re-implementation of this feature on master.

PR #1855, opened 2020-01-22
fix: update the java out type mapping of std::vector<CType*>&

andrew-teirney:
This change attempts to resolve the cause of segfault(s) that occur when retrieving the break intervals of a vehicle for a RoutingDimension using the Java SWIG bindings. 

```java
package tmp;

import com.google.ortools.constraintsolver.RoutingDimension;
import com.google.ortools.constraintsolver.RoutingIndexManager;
import com.google.ortools.constraintsolver.RoutingModel;

public class Tmp {
    static {
        System.loadLibrary("jniortools");
    }

    public static void main(String[] args) {
        RoutingIndexManager manager = new RoutingIndexManager(2, 1, new int[]{0}, new int[]{1});
        RoutingModel model = new RoutingModel(manager);
        model.addConstantDimension(0, 0, false, "tmp");
        RoutingDimension dimension = model.getMutableDimension("tmp");
        // The line below causes a segfault 
        dimension.getBreakIntervalsOfVehicle(0);
    }
}
```

### DIFF
--- a/ortools/util/java/vector.i
+++ b/ortools/util/java/vector.i
@@ -132,16 +132,19 @@ VECTOR_AS_JAVA_ARRAY(double, double, Double);
   $1 = &result;
 }
 %typemap(out) const std::vector<CType*>& {
-  jclass object_class = jenv->FindClass("JavaPackage/JavaType");
+  if (nullptr == $1)
+    return $null;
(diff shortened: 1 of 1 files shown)


lperron:
Thanks

Mizux:
LGTM

PR #1938, opened 2020-03-26
Add uncapacitated facility location example

daespino:


daespino:
@googlebot I signed it!

PR #1956, opened 2020-04-07
[Python] store start, end, size in IntervalVar

stradivari96:
A common pattern I see when using IntervalVars is that you usually want to keep track of its `start`, `end` and `size` (I guess also the `is_present` literal).

See:
https://github.com/google/or-tools/blob/858fa626959f7e386153af82756384b79f983b5a/examples/python/jobshop_ft06_distance_sat.py#L58-L70

Can you consider adding them as attributes of the IntervalVar class?

### DIFF
--- a/ortools/sat/python/cp_model.py
+++ b/ortools/sat/python/cp_model.py
@@ -642,7 +642,7 @@ class IntervalVar(object):
   """
 
     def __init__(self, model, start_index, size_index, end_index,
-                 is_present_index, name):
+                 is_present_index, name, start=None, end=None, size=None):
         self.__model = model
         self.__index = len(model.constraints)
         self.__ct = self.__model.constraints.add()
@@ -653,6 +653,10 @@ class IntervalVar(object):
             self.__ct.enforcement_literal.append(is_present_index)
         if name:
             self.__ct.name = name
+
+        self.__start = start
+        self.__end = end
+        self.__size = size
 
     def Index(self):
         """Returns the index of the interval constraint in the model."""
@@ -682,6 +686,14 @@ class IntervalVar(object):
     def Name(self):
         return self.__ct.name
 
+    def Start(self):
+        return self.__start
+
+    def End(self):
+        return self.__end
+
+    def Size(self):
+        return self.__size
 
 class CpModel(object):
     """Methods for building a CP model.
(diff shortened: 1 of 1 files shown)


ortools/sat/python/cp_model.py
```diff
@@ -654,6 +654,10 @@ def __init__(self, model, start_index, size_index, end_index,
         if name:
             self.__ct.name = name
 
+        self.start = None
+        self.end = None
+        self.size = None
```
Mizux, line 659: shouldn't be self.__start = None + def Start(self):... ? for consistency

ortools/sat/python/cp_model.py
```diff
@@ -654,6 +654,10 @@ def __init__(self, model, start_index, size_index, end_index,
         if name:
             self.__ct.name = name
 
+        self.__start = start
```
lperron, line 657: Sorry, I have to disagree with this change. As you can change the proto behind the back, I would prefer recreating the IntVar on demand.

ortools/sat/python/cp_model.py
```diff
@@ -682,6 +686,14 @@ def __repr__(self):
     def Name(self):
         return self.__ct.name
 
+    def Start(self):
```
lperron, line 689: I would do: def Start(self, model):
  return IntVar(model, self.__ct.interval.start)

with a new IntVar ctor

ortools/sat/python/cp_model.py
stradivari96: True, I didn't think about a possible mismatch and this is quite easy to do outside the library anyway.

ortools/sat/python/cp_model.py
stradivari96: Ups, I misunderstood your earlier comment, I thought you disagreed with the whole idea. I will correct this later

stradivari96: One thing that could be troublesome is that this will create a new object, so if someone expects the same object for things like dictionaries, sets and so on, it wont work. Unless we also redefine `__hash__` with the index or something

lperron:
I looked at it. It does not work :-(

Please revert the PR.
00

stradivari96:
Yeah, I'll leave this closed for now (wasn't merged), thanks for your time!

PR #1967, opened 2020-04-19
Support management of visit types using the Java SWIG interface

andrew-teirney:
This changeset exposes the methods necessary to support for management of visit types and adding temporary and hard type compatibilities via the Java SWIG interface.


Mizux:
IIRC I ignored these methods since the VisitType is not correctly typemaped so you'll have tons of SWIG_PTR_.... ugly opaque pointer type.

Need to double check...

andrew-teirney:
Based on my review of the methods the visit types are just `int`.

PR #1969, opened 2020-04-20
Expose packCumulsOfOptimizerDimensionsFromAssignment methods to Java

andrew-teirney:
This changeset provides to the java language the ability to call the C++ PackCumulsOfOptimizerDimensionsFromAssignment methods on RoutingModel.

There are two methods exposed, one that takes a timeout in milliseconds and another that does not and will, therefore, have an infinite timeout.


andrew-teirney:
This pull request can be closed without merging as I no longer need access to these methods.

PR #1998, opened 2020-05-01
[CBC] Use setBestSolution to set hint

stradivari96:
Hi, I'm still testing this but could you take a quick look to see if this is going in the right direction?
Also, I am new to c++, so sorry if there is any simple language usage mistake.

Thanks

### DIFF
--- a/ortools/linear_solver/cbc_interface.cc
+++ b/ortools/linear_solver/cbc_interface.cc
@@ -349,6 +349,23 @@ MPSolver::ResultStatus CBCInterface::Solve(const MPSolverParameters& param) {
   // Solve
   CbcModel model(osi_);
 
+  // Set solution hints if any
+  if (!solver_->solution_hint_.empty()) {
+    const int num_vars = solver_->variables_.size();
+    if (solver_->solution_hint_.size() != num_vars) {
+      VLOG(1) << "Ignoring partial solution hint.";
+    } else {
+      // +1 as CBC adds a "dummy" variable with index 0
+      std::vector<double> hint_values(num_vars+1);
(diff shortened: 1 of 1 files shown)


ortools/linear_solver/cbc_interface.cc
```diff
@@ -349,6 +349,23 @@ MPSolver::ResultStatus CBCInterface::Solve(const MPSolverParameters& param) {
...
+      VLOG(1) << "Ignoring partial solution hint.";
+    } else {
+      // +1 as CBC adds a "dummy" variable with index 0
+      std::vector<double> hint_values(num_vars+1);
+      for (const std::pair<const MPVariable*, double>& p :
+         solver_->solution_hint_) {
+        hint_values[MPSolverVarIndexToCbcVarIndex(p.first->index())] = p.second;
+      }
+      // Set objectiveValue as COIN_DBL_MAX as we do not care
+      model.setBestSolution(&hint_values[0], num_vars+1, COIN_DBL_MAX, true);
```
Mizux, line 365: I would use: 
```cpp
model.setBestSolution(hint_values.data(), hint_values.size(), COIN_DBL_MAX, /*check=*/true)
```
note: At Google for boolean flag, it seems the convention is to add the parameter name as hint concerning the meaning of the boolean in the call site.

ref: https://github.com/coin-or/Cbc/blob/7b5ccc016f035f56614c8018b20d700978144e9f/Cbc/src/CbcModel.hpp#L1471-L1478

stradivari96: Yeah, I should take a look at the c++ style guide...
Let me add a test and try the CBC behavior a bit more before merging

PR #2024, opened 2020-05-18
Remove redundant CapInt64

stradivari96:
There is no need to Cap because it is handled by:
* `AssertIsInt64(arg)` (arg in [INT_MIN, INT_MAX])
* `if arg == INT_MAX`
* `if arg == INT_MIN`.

https://github.com/google/or-tools/blob/532635de74c83d619fe957b3fd0eca1e94ac5b99/ortools/sat/python/cp_model.py#L284-L288

https://github.com/google/or-tools/blob/532635de74c83d619fe957b3fd0eca1e94ac5b99/ortools/sat/python/cp_model.py#L294-L298

https://github.com/google/or-tools/blob/532635de74c83d619fe957b3fd0eca1e94ac5b99/ortools/sat/python/cp_model.py#L306-L316

PS: Will try to find a more tangible and meaninful improvement in the next PR...

### DIFF
--- a/ortools/sat/python/cp_model.py
+++ b/ortools/sat/python/cp_model.py
@@ -285,7 +285,7 @@ class LinearExpr(object):
             if arg == INT_MIN:
                 raise ArithmeticError('< INT_MIN is not supported')
             return BoundedLinearExpression(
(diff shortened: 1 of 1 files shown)


lperron:
Nice!

PR #2095, opened 2020-07-06
Add ABSL_INC to DEPENDENCIES_INC

jspricke:


Mizux:
Thanks for this patch !

But I will closed it since:
- our dev branch is `master` not `stable` (release branch)
- we should also do it for the `SWIG_INC` flags too
- I think i'll need to rework it, like we did for DEPENDENCIES_LNK i.e. move it in third_party.unix

PR #2104, opened 2020-07-15
Help CMake resolve protobuf::protoc when not building it.

mikepurvis:
I required this change to build or-tools using cmake with the `-DBUILD_DEPS=0` option. Without this change, I get this at build time:

```
CMakeFiles/ortools_proto.dir/build.make:61: *** target pattern contains no '%'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:590: CMakeFiles/ortools_proto.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```


Mizux:
Thx for the patch !

warning: currently all FindXXX are more or less "placeholder/blind fix" since we only test with `-DBUILD_DEPS=ON`, but I want to support system wide dependencies in the near future so currently your ahead of our development ;)

PR #2107, opened 2020-07-16
In CMake, install python to install prefix.

mikepurvis:
Currently the CMake build prepares the `build/python` directory as a source Python package, and by default does a test install of it into `build/venv` when `BUILD_TESTING` is enabled.

This change enables a conventional `cmake ..; make; sudo make install` to actually install the package into the specified prefix (default `/usr/local`), where it will be seen by the system python. This also needed to support conventional packaging workflows where the DESTDIR envvar is specified at install time.

### DIFF
--- a/cmake/python-install.cmake.in
+++ b/cmake/python-install.cmake.in
@@ -1,0 +1,7 @@
+execute_process(
+  WORKING_DIRECTORY @PROJECT_BINARY_DIR@/python
+  COMMAND @Python_EXECUTABLE@ setup.py install
+    --root=/
+    --prefix=@CMAKE_INSTALL_PREFIX@
+    @SETUPTOOLS_INSTALL_LAYOUT@
+)
--- a/cmake/python-install.cmake.in
+++ b/cmake/python-install.cmake.in
@@ -1,7 +1,13 @@
+if (DEFINED ENV{DESTDIR})
+  set(ROOT $ENV{DESTDIR})
+else()
+  set(ROOT /)
+endif()
+
 execute_process(
   WORKING_DIRECTORY @PROJECT_BINARY_DIR@/python
   COMMAND @Python_EXECUTABLE@ setup.py install
-    --root=/
+    --root=${ROOT}
     --prefix=@CMAKE_INSTALL_PREFIX@
     @SETUPTOOLS_INSTALL_LAYOUT@
 )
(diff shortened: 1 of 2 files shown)


Mizux:
thanks for this PR,

Actually I was thinking to an alternative.
- not sure `cmake --build build --target install` is the good candidate to install the python (i.e. using `install(SCRIPT)` ) since you'll install both the C++ "classic" stuff as well as the python "package", while most of our python user just want to ONLY install the python package
- Like we did for the makefile, I'll prefer to have a standalone `custom_target` (ed not in ALL) let's call it "install_python" which will only install the python package. 
- Prefer to perform a "user" install than a global one.
- Actually I'm thinking on using `"$(Python_EXECUTABLE)" setup.py install --user` or `"$(Python_EXECUTABLE)" -m pip install --user --find-links=${CMAKE_CURRENT_BINARY_DIR}/python/dist ortools`

- concerning `dist-package` / `site-package`, thanks for bringing me this issue but It feels strange to me that it is not already take into account by the python system wide installation directly, ie if debian folk want to change the default install layout they should adapt the python setuptools package default behaviour accordingly IMHO
Also is it the case for user install aka `pip install --user` ? need to dig this subject...

mikepurvis:
This PR is aligned to what libraries like OpenCV do with their libs and bindings when your `cmake; make install` them, and it's what works for use with workspace-oriented build helpers like `catkin_tools` and `colcon` (from the ROS ecosystem). Directly installing into `CMAKE_PREFIX_PATH` with the `DESTDIR` envvar supported is also what's needed to create deb/rpm packages with minimal fuss.

Certainly if you're going to do a `--user` install, that definitely shouldn't be part of ALL as it would be quite surprising to most users, especially if they expecting an installation into {{/usr/local}} and therefore invoking the target with {{sudo}}.

Anyway, these use cases may not be your primary focus, but that's what motivates the change being as it is.

jmarca:
my 2 cents:  as a python user, I agree that a --user install by default is surprising and unexpected.  I would prefer the current python makefiles change to the behavior proposed in this cmake patch.  

I actually deleted the --user flag from my local version of the python make file (Makefile.python.mk) as it was installing to the *root* user's local library in my docker image.

Mizux:
FYI I rework it here with few change
* don't set root to / when DESTDIR is not used (also i didn't think it was working on windows BTW)
* if DESTDIR is a relative path, make it relative to PROJECT_BINARY_DIR since it is the the way install(TARGET ...) work
  note: In your PR it was relative to PROJECT_BINARY_DIR/python so if using DESTDIR=tmp we have c++ stuff installed in `<build_dir>foo/usr/local/...` while python was installed in `<build_dir>/python/foo/usr/local/...`

PR #2165, opened 2020-09-24
Change to new MiniZinc library structure

guidotack:
This PR updates the MiniZinc library to the new structure, where global constraints are defined in `fzn_` files. It also makes the following changes:

- Constraints `table_int`, `table_bool`, `regular` and `network_flow` are now converting the 2d array arguments to 1d  arrays, so that the generated FlatZinc is in fact valid MiniZinc.
- Removed some legacy files (`table.mzn`, the `minimum`/`maximum` files) that should not be redefined in a solver library

I noticed that there is a definition for `table_bool`, however this constraint does not seem to be handled by the c++ code, so this may have to be fixed as well.


lperron:
Thanks!

PR #2250, opened 2020-11-30
Feature/routing breaks example

newmonade:
This is a new routing example. It illustrate how to set breaks for more than one vehicle, using IntVar to represent the beginning of the break. As an example, it allows to define breaks as "30 minute break starting between 3 and 4h after the route starts".

I think it illustrate the request: https://stackoverflow.com/questions/63816576/google-or-tools-driver-breaks-for-hours-of-service-regulations.


newmonade:
@googlebot I signed it!

Mizux:
@googlebot I consent.

PR #2251, opened 2020-12-01
Expose routing API for node types and breaks in Java

newmonade:
This PR aims at exposing the C++/Python routing API about node types and breaks in Java.


ortools/constraint_solver/routing.h
```diff
@@ -2550,12 +2550,9 @@ class RoutingDimension {
   /// never we called on any (end, _). If pre_travel_evaluator or
   /// post_travel_evaluator is -1, it will be taken as a function that always
   /// returns 0.
-  // TODO(user): Remove if !defined when routing.i is repaired.
-#if !defined(SWIGPYTHON)
-  void SetBreakIntervalsOfVehicle(std::vector<IntervalVar*> breaks, int vehicle,
+  void wSetBreakIntervalsOfVehicle(std::vector<IntervalVar*> breaks, int vehicle,
```
lperron, line 2553: not wSetBreak...

ortools/constraint_solver/java/routing.i
```diff
@@ -102,18 +98,11 @@ import java.util.function.LongUnaryOperator;
 %ignore RoutingModel::PackCumulsOfOptimizerDimensionsFromAssignment;
 %ignore RoutingModel::RegisterStateDependentTransitCallback;
 %ignore RoutingModel::RemainingTime;
-%ignore RoutingModel::SolveWithParameters(
-    const RoutingSearchParameters& search_parameters,
-    std::vector<const Assignment*>* solutions);
```
lperron, line 107: Is this supported ? (the vector* as parameter)

Mizux: IIRC no, that's why it is currently not wrapped because it will lead to tons of swig PINVOKE opaque pointer making this API ugly and unusable.

ortools/constraint_solver/routing.h
newmonade: yes that was a typo

ortools/constraint_solver/java/routing.i
newmonade: I will revert to ignore that one and SolveFromAssignmentWithParameters with the vector<const Assignment*>* solutions argument as well

Mizux:
Seems to be a revamp of #1967

ortools/constraint_solver/routing.h
```diff
@@ -2550,12 +2550,9 @@ class RoutingDimension {
   /// never we called on any (end, _). If pre_travel_evaluator or
   /// post_travel_evaluator is -1, it will be taken as a function that always
   /// returns 0.
-  // TODO(user): Remove if !defined when routing.i is repaired.
-#if !defined(SWIGPYTHON)
```
Mizux, line 2554: remove this you also allow it in .Net wrapper, not sure you have checked it...
On my way to do some tests...

Mizux: My bad this is the contrary, it unignores it in Python contrary to previous behaviour and it make examples failing...

PR #2291, opened 2020-12-20
Add extra braces around comparison for readability

agentydragon:


Mizux:
note: already backported in Google codebase (@agentydragon you should have been in copy of the CL)

PR #2356, opened 2021-01-24
Vehicle routing problem with fixed penalty time windows

owenlacey:
I couldn't find any examples like this online, so thought I'd create an example, as I've had this problem to solve recently.

This is essentially a vehicle routing problem with the following constraints:
- Each order has a time window
- Rather than enforcing the model conforming to the time windows, we'll penalise any lateness.
- This penalty with be fixed, as opposed to a progressively worse penalty depending on how late it is.

### DIFF
--- a/examples/contrib/vrptw_fixed_penalty.cs
+++ b/examples/contrib/vrptw_fixed_penalty.cs
@@ -1,4 +1,4 @@
-// Copyright 2012 Hakan Kjellerstrand
+// Copyright 2021 Owen Lacey
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/examples/contrib/vrptw_fixed_penalty.cs
+++ b/examples/contrib/vrptw_fixed_penalty.cs
@@ -1,0 +1,169 @@
+// Copyright 2012 Hakan Kjellerstrand
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


owenlacey:
@googlebot I fixed it.

owenlacey:
@googlebot I fixed it.

examples/contrib/vrptw_fixed_penalty.cs
```diff
@@ -0,0 +1,169 @@
+// Copyright 2012 Hakan Kjellerstrand
```
Mizux, line 1: Not sure of this copyright claims, should be this instead
`// Copyright 2021 owenlacey` (or both lines ?)
or whatever you want (e.g. your employer if any).

Mizux (approved):

Mizux:
note: PR must be done against our dev branch which is master, stable branch is our default and release branch.

ps: Github force default branch (for cloning, and on github.com) to be also the default branch for PR -_-

examples/contrib/vrptw_fixed_penalty.cs
owenlacey: of course - sorry that's some lazy copy-pasting on my part

stradivari96:
Also make sure you signed the CLA with the email that appears here:
https://github.com/google/or-tools/commit/e5d09b54aeb603acadab7d3da10ac9eafbf8926f.patch

owenlacey:
Apologies, I'm going to abandon this because the commits were authored by an alternative (and incorrect account). I couldn't see a quick way around this so thought I'd redo the changes in another PR.

Other pr here: https://github.com/google/or-tools/pull/2358

PR #2358, opened 2021-01-25
Vehicle routing problem with fixed penalty time windows (re-make of #2356)

owenlacey:
NB this is a replica of https://github.com/google/or-tools/pull/2356, which I unfortunately had to redo because the commits were authored by the wrong email address (for CLA purposes).

I couldn't find any examples like this online, so thought I'd create an example, as I've had this problem to solve recently.

This is essentially a vehicle routing problem with the following constraints:
- Each order has a time window
- Rather than enforcing the model conforming to the time windows, we'll penalise any lateness.
- This penalty with be fixed, as opposed to a progressively worse penalty depending on how late it is.

### DIFF
--- a/examples/contrib/vrptw_fixed_penalty.cs
+++ b/examples/contrib/vrptw_fixed_penalty.cs
@@ -1,0 +1,169 @@
+// Copyright 2021 Owen Lacey
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Mizux (approved):

Mizux:
waiting for few .Net jobs to run then should be merged ASAP.

Thx !

PR #2382, opened 2021-02-04
Wrap CP-SAT assumptions

stradivari96:
Some questions:
- Is this the desired API or maybe `model.AddAssumptions([var1, var2])`?
- Not sure if returning the indices of the boolvars is very user-friendly but saving the variables in the solver also seems like a bad idea, are there any other good alternatives?

### DIFF
--- a/ortools/java/com/google/ortools/sat/CpModel.java
+++ b/ortools/java/com/google/ortools/sat/CpModel.java
@@ -966,6 +966,18 @@ public final class CpModel {
     modelBuilder.getSolutionHintBuilder().addValues(value);
   }
 
+  /** Adds variable as assumption */
+  public void addAssumption(Literal lit) {
+    modelBuilder.addAssumptions(lit.getIndex());
+  }
+
+  /** Adds multiple variables to the assumptions */
+  public void addAssumptions(Literal[] literals) {
+    for (Literal lit : literals) {
+      addAssumption(lit);
+    }
+  }
+
   // Objective.
 
   /** Adds a minimization objective of a linear expression. */
--- a/ortools/java/com/google/ortools/sat/CpModel.java
+++ b/ortools/java/com/google/ortools/sat/CpModel.java
@@ -966,6 +966,11 @@ public final class CpModel {
     modelBuilder.getSolutionHintBuilder().addValues(value);
   }
 
+  /** Remove all solution hints */
+  public void clearHints() {
+    modelBuilder.clearSolutionHint();
+  }
+
   /** Adds variable as assumption */
   public void addAssumption(Literal lit) {
     modelBuilder.addAssumptions(lit.getIndex());
@@ -976,6 +981,11 @@ public final class CpModel {
     for (Literal lit : literals) {
       addAssumption(lit);
     }
+  }
+
+  /** Remove all solution hints */
+  public void clearAssumptions() {
+    modelBuilder.clearAssumptions();
   }
 
   // Objective.
--- a/ortools/sat/cp_model.cc
+++ b/ortools/sat/cp_model.cc
@@ -796,6 +796,16 @@ void CpModelBuilder::AddHint(IntVar var, int64 value) {
   cp_model_.mutable_solution_hint()->add_values(value);
 }
 
+void CpModelBuilder::AddAssumption(BoolVar lit) {
+  cp_model_.mutable_assumptions()->Add(lit.index_);
+}
+
+void CpModelBuilder::AddAssumptions(absl::Span<const BoolVar> literals) {
+  for (const BoolVar& lit : literals) {
+    cp_model_.mutable_assumptions()->Add(lit.index_);
+  }
+}
+
 int64 SolutionIntegerValue(const CpSolverResponse& r, const LinearExpr& expr) {
   int64 result = expr.constant();
   for (int i = 0; i < expr.variables().size(); ++i) {
--- a/ortools/sat/cp_model.cc
+++ b/ortools/sat/cp_model.cc
@@ -796,6 +796,10 @@ void CpModelBuilder::AddHint(IntVar var, int64 value) {
   cp_model_.mutable_solution_hint()->add_values(value);
 }
 
+void CpModelBuilder::ClearHints() {
+  cp_model_.mutable_solution_hint()->Clear();
+}
+
 void CpModelBuilder::AddAssumption(BoolVar lit) {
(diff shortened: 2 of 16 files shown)


ortools/sat/csharp/CpModel.cs
```diff
@@ -615,6 +615,11 @@ public void AddHint(IntVar var, long value)
             model_.SolutionHint.Values.Add(value);
         }
 
+        public void AddAssumption(IntVar var)
```
lperron, line 618: Wrong type. Maybe we should accept a iLiteral, and make sure the literal is positive.

ortools/sat/python/cp_model.py
```diff
@@ -1542,6 +1542,9 @@ def AddHint(self, var, value):
         self.__model.solution_hint.vars.append(self.GetOrMakeIndex(var))
```
lperron, line 1542: We need to make sure it is positive

ortools/sat/samples/assumptions_sample_sat.cc
```diff
@@ -0,0 +1,59 @@
...
+  const BoolVar b = cp_model.NewBoolVar().WithName("b");
+  // [END variables]
+
+  // [START constraints]
+  cp_model.AddGreaterThan(x, 10).OnlyEnforceIf(a);
+  cp_model.AddLessOrEqual(y, 10).OnlyEnforceIf(b);
+  // [END constraints]
+
+  // Add assumptions
+  cp_model.AddAssumption(a);
```
lperron, line 40: You are right, it should be nice to have a vector/list/enumerable API.

ortools/java/com/google/ortools/sat/CpModel.java
```diff
@@ -966,6 +966,13 @@ public void addHint(IntVar var, long value) {
     modelBuilder.getSolutionHintBuilder().addValues(value);
   }
 
+  /** Adds variable as assumption */
+  public void addAssumptions(Literal[] literals) {
+    for (Literal lit : literals) {
+      modelBuilder.addAssumptions(lit.getIndex());
+    }
+  }
```
lperron, line 974: I would add both the single var, and the var[] API.

stradivari96:
Thanks for the fast review! I have added the proposed changes

ortools/sat/cp_model.cc
```diff
@@ -796,6 +796,12 @@ void CpModelBuilder::AddHint(IntVar var, int64 value) {
   cp_model_.mutable_solution_hint()->add_values(value);
 }
 
+void CpModelBuilder::AddAssumptions(absl::Span<const BoolVar> literals) {
+  for (const BoolVar& lit : literals) {
```
lperron, line 800: Single var API and this one. Thanks

ortools/java/com/google/ortools/sat/CpModel.java
stradivari96: Okey!

ortools/sat/cp_model.cc
```diff
@@ -796,6 +796,12 @@ void CpModelBuilder::AddHint(IntVar var, int64 value) {
...
+    cp_model_.mutable_assumptions()->Add(lit.index_);
```
lperron, line 801: ->Add(PositiveRef(lit.index_));

ortools/java/com/google/ortools/sat/CpModel.java
stradivari96: Overload or two methods (without the s)?

ortools/sat/csharp/CpModel.cs
```diff
@@ -615,6 +615,14 @@ public void AddHint(IntVar var, long value)
...
         }
 
+        public void AddAssumptions(IEnumerable<ILiteral> literals)
+        {
+            foreach (ILiteral lit in literals)
+            {
+                model_.Assumptions.Add(lit.GetIndex());
+            }
+        }
+
```
lperron, line 625: add single var API, checks Positivity: 
if (lit.GetIndex() >= 0) {
  model_.Assumptions.Add(lit.GetIndex());
} else {
  model_.Assumptions.Add(lit.Not().GetIndex());
}

to be done in python, java, and .NET

lperron:
Thanks for doing this.

lperron:
2 methods, 1 with S, 1 without

00



Le jeu. 4 févr. 2021 à 20:30, Xiang Chen <notifications@github.com> a
écrit :

ortools/sat/csharp/CpModel.cs
stradivari96: Doesn't this change the semantics for a (nonsensical) case like this?
```python
x = model.NewIntVar(0, 10, "")
a = model.NewBoolVar("")

model.Add(x > 10).OnlyEnforceIf(a.Not())
model.Proto().assumptions.append(a.Not().Index())
```
How about raise an error instead?

ortools/sat/python/cp_model.py
```diff
@@ -1544,8 +1544,17 @@ def AddHint(self, var, value):
...
+        self.__model.ClearField("solution_hint")
+
+    def AddAssumption(self, lit):
+        self.__model.assumptions.append(self.GetOrMakeBooleanIndex(lit))
+
+    def AddAssumptions(self, literals):
+        for lit in literals:
+            self.AddAssumption(lit)
+
+    def ClearAssumptions(self):
```
lperron, line 1556: I suppose we need this method in all languages.

lperron:
Thanks for the effort.

two things: 
for the example, I would do the following:
3 vars (x, y, z), 3 literals (a, b, c). a => x > y, b => y > z, c => z > x. All 3 literals are assumptions.

lperron:
second thing, returning the index raises a lot of question.  Keep it that way for now. I will revisit this.

stradivari96:
Almost there! I think the only remaining thing is my question about Positive Refs.

lperron:
Let me merge and tweak it further.

Thanks!!!

stradivari96:
Thanks for your guidance!

lperron:
It turns out the model accepts positive and negative literals as
assumptions.

I will push some minor fixes and start looking at reading the result from
the model.
00



Le ven. 5 févr. 2021 à 17:03, Xiang Chen <notifications@github.com> a
écrit :

stradivari96:
In case it helps, as an user I think something like `solver.ValidAssumption(lit) -> bool` should be good, not sure what to do if you return multiples unsat cores though.

PR #2403, opened 2021-02-20
Enable Emscripten builds

kjartanm:
To enable Emscripten builds, we need to fence off a couple of elements that creates unrecoverable errors during build process.


kjartanm:
Will create a new against master

Issue #1670, opened 2019-10-25
Feature request: A browser-friendly version of or-tools


franzen:
With web development frameworks such as Vue, React and Angular, a web-based version (WASM) of or-tools would make a big difference for us developers. The ability to solve many optimization problems directly in the browser would provide a better user experience and the opportunity to do work in the browser when offline.

Unfortunately, I'm not a star on tools like https://emscripten.org/, otherwise I would have made the first attempt myself. But can it be a simple task for someone else in this community?

Thank you for your hard work on this project!

lperron:
As expected, you, or somebody else can try.  :-)
 But I will not spend some time on it. I prefer the client server approach.

--Laurent

Le ven. 25 oct. 2019 à 22:33, Nils Franzen <notifications@github.com> a
écrit :

CervEdin:
There's a reason solvers and local search algorithms aren't implemented in high level languages. They're not suited for solving NP-hard problems.

If you really need to do this client side, you could look into packages like https://www.npmjs.com/package/simulated-annealing or something similar.

If you just need a simpler interface for modelling optimization problems client-side, check out the MiniZinc project.

kjartanm:
An experimental/poc wasm-version of or-tools is available at https://github.com/kjartanm/wasm-or-tools (and wasm is not exactly high level :) )

Mizux:
Copy of the mailing list topic: **Proof of concept WASM-build of OR-Tools**

Few comments for myself (and sorry, I should have done this feedback  20 days ago when you reply in the issue):
* Build on top of stable branch (v7.8)
https://github.com/kjartanm/wasm-or-tools/commits/stable

* Basically use the CMake based build with an emscripten CMAKE_TOOLCHAIN_FILE -> GOOD
ed: While we still not support cmake toolchain files (i.e. not tested and out of our current scope), it is the way to go to support Android NDK and emscripten in the near future..
So any clean patch (i.e. without breaking change) to ease this future usage will be reviewed and integrated if possible !

* Have some dirty hack to "fix" dependencies (glog and protobuf)
https://github.com/kjartanm/wasm-or-tools/commit/ef4b4fadea2c9550c0e81890490d844281ef65c7 
note: For Glog why did we need to use "HAVE_SYMBOLIZE 0" ? and the hack of raw_looging -> provide a clean patch to glog to better support toolchain and/or emscripten target should be better IMHO...
[...]

PR #2404, opened 2021-02-20
Add Wasm32 as target for cross compiling

kjartanm:
This PR adds Wasm32 as target for cross compiling, Emscripten setup,  and fences off some Emscripten-incompatible elements.

See Feature request #1670

### DIFF
--- a/ortools/base/raw_logging.cc
+++ b/ortools/base/raw_logging.cc
@@ -25,13 +25,17 @@
 #include "ortools/base/logging.h"
 #include "ortools/base/logging_utilities.h"
 
-#if !defined(_MSC_VER)
+#if !defined(_MSC_VER) && !defined(__EMSCRIPTEN__)
 #include <sys/syscall.h>  // for syscall()
 #define safe_write(fd, s, len) syscall(SYS_write, fd, s, len)
 #else
+#if !defined(__EMSCRIPTEN__)
 #include <io.h>  // _write()
 // Not so safe, but what can you do?
 #define safe_write(fd, s, len) _write(fd, s, len)
+#else
+#define safe_write(fd, s, len) write(fd, s, len)
+#endif
 #endif
 
 #if defined(_MSC_VER) && !defined(__MINGW32__)
--- a/tools/cross_compile.sh
+++ b/tools/cross_compile.sh
@@ -95,6 +95,19 @@ function clean_build() {
   mkdir -p "${BUILD_DIR}"
 }
 
+function expand_wasm_config() {
+  declare -r EMSDK_VERSION=2.0.14
+  declare -r EMSDK_URL=https://github.com/emscripten-core/emsdk/archive/${EMSDK_VERSION}.tar.gz 
+  declare -r EMSDK_RELATIVE_DIR="emsdk-${EMSDK_VERSION}"
+  declare -r EMSDK="${ARCHIVE_DIR}/${EMSDK_RELATIVE_DIR}"
+  if [ ! -d "${EMSDK}" ]; then
+    install_emscripten
+  fi
+
+  declare -r TOOLCHAIN_FILE=${EMSDK}/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake
+  CMAKE_ADDITIONAL_ARGS+=( -DCMAKE_TOOLCHAIN_FILE="${TOOLCHAIN_FILE}" -DBUILD_SAMPLES=OFF -DBUILD_EXAMPLES=OFF -DBUILD_TESTING=OFF)
+}
+
 function expand_linaro_config() {
   #ref: https://releases.linaro.org/components/toolchain/binaries/
   local -r LINARO_VERSION=7.5-2019.12
@@ -322,6 +335,9 @@ function main() {
     mips64el)
       expand_codescape_config
       declare -r QEMU_ARCH=mips64el ;;
+    wasm32)
+      expand_wasm_config
+      declare -r QEMU_ARCH=DISABLED ;;
     *)
       >&2 echo "Unknown TARGET '${TARGET}'..."
       exit 1 ;;
--- a/ortools/glop/lp_solver.cc
+++ b/ortools/glop/lp_solver.cc
@@ -160,6 +160,7 @@ ProblemStatus LPSolver::SolveWithTimeLimit(const LinearProgram& lp,
   // Note that we only activate the floating-point exceptions after we are sure
   // that the program is valid. This way, if we have input NaNs, we will not
   // crash.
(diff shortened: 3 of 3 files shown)


ortools/base/raw_logging.cc
```diff
@@ -25,13 +25,17 @@
 #include "ortools/base/logging.h"
 #include "ortools/base/logging_utilities.h"
 
-#if !defined(_MSC_VER)
+#if !defined(_MSC_VER) && !defined(__EMSCRIPTEN__)
 #include <sys/syscall.h>  // for syscall()
 #define safe_write(fd, s, len) syscall(SYS_write, fd, s, len)
 #else
+#if !defined(__EMSCRIPTEN__)
```
lperron, line 32: Can you reverse the test

#if defined(__EMSCRIPTEN__)

#else

kjartanm: Have changed the test.

maldito-dot (commented):
@maldito-dot

tools/cross_compile.sh
```diff
@@ -95,6 +95,25 @@ function clean_build() {
...
+function expand_wasm_config() {
+  local -r EMSDK_VERSION=2.0.14
+  local -r EMSDK_URL=https://github.com/emscripten-core/emsdk/archive/${EMSDK_VERSION}.tar.gz 
+  local -r EMSDK_RELATIVE_DIR="emsdk-${EMSDK_VERSION}"
+  local -r EMSDK="${ARCHIVE_DIR}/${EMSDK_RELATIVE_DIR}"
+  if [ ! -d "${EMSDK}" ]; then
+    echo "Fetching emscripten"
+    unpack "${EMSDK_URL}" "${EMSDK_RELATIVE_DIR}"
+    echo "Installing Emscripten ..."
+    ${EMSDK}/emsdk install ${EMSDK_VERSION}
```
Mizux, line 107: need to check what 'install' cmd do...

Mizux (approved):

tools/cross_compile.sh
kjartanm: It just installs emsdk locally.

akutruff:
What's blocking WASM support right now?  Is this dead?

Mizux:
mostly lack human ressource (i.e. lack of time) to work on it + fear of the technical debt increase to maintain it.

BTW I'm not against it, on the contrary I'd love to have an ortools-lite running on the edge on client side web browser...

Mizux (approved):

akutruff:
Wish I could help on the build errors - Seems like there's some simple permissions error somewhere?

chadbrewbaker:
I have an immediate need for this. What is the blocker?

This is how Z3 does it, they provide not just a [ WASM build](https://github.com/Z3Prover/z3/blob/master/src/api/js/scripts/build-wasm.ts) but also a JS interface. I agree this should be effectively an or-tools-lite in the default build, you should have to pass some extra flag to build a really fat binary.

Minimal viable commit is defining a JSON for an or-tools graph format and having it pass some unit tests?

Mizux:
My 2 cents,

We have a job to test emscripten build
https://github.com/google/or-tools/actions/workflows/amd64_web.yml
src: https://github.com/google/or-tools/blob/stable/.github/workflows/amd64_web.yml

using:
`make --directory=cmake web_build`
src: https://github.com/google/or-tools/blob/ef7bb07d57bbe73c5e2ff5bcff9894e0e3164f56/cmake/Makefile#L660-L698

Dockerfile: https://github.com/google/or-tools/blob/stable/cmake/docker/web/Dockerfile

Need to compare/test with this PR...

tools/cross_compile.sh
```diff
@@ -265,6 +265,25 @@ QEMU_ARGS+=( -L "${SYSROOT_DIR}" )
 QEMU_ARGS+=( -E LD_PRELOAD="${SYSROOT_DIR}/usr/lib/libstdc++.so.6:${SYSROOT_DIR}/lib/libgcc_s.so.1" )
 }
 
+function expand_wasm_config() {
+  local -r EMSDK_VERSION=2.0.14
```
Mizux, line 269: Last Tag as of 2024 June 07
https://github.com/emscripten-core/emsdk/releases/tag/3.1.61

tools/cross_compile.sh
```diff
@@ -403,6 +423,9 @@ function main() {
     ppc)
       expand_bootlin_config
       declare -r QEMU_ARCH=ppc ;;
+    wasm32)
```
Mizux, line 426: note for myself, emscripten generate wasm32 by default
ref: https://emscripten.org/docs/tools_reference/settings_reference.html#memory64

tools/cross_compile.sh
airen1986: Hi Mizux,
Is it planned for next release or will it take time? Also, is this command work with stable branch:
emcmake cmake -S . -B build -DBUILD_DEPS=ON -DBUILD_PYTHON=ON

I tried after making changes in cross_compile.sh and I am getting error as:
Could NOT find Python3 (missing: Python3_INCLUDE_DIRS Python3_LIBRARIES Development Development.Module Development.Embed) (found version "3.11.9")

However, I don't see such error when using just cmake?

Mizux: Sorry, I don't get it, what did you try to do/expect ?

`BUILD_PYTHON` is used to generate automatically python wrapper on top of the C++ library to have access to ortools in python (which is just a wrapper on top of the native library, since we won't reimplement ortools in pure Python/Java nor .Net).

On the other side, AFAIK, `emcmake` is used to transpile C++ code into wasm byte code to have a "browser compatible" source code so there is no point to try to build a python wrapper on top of this wasm library.

agriyakhetarpal: Hi there, I recently pointed @airen1986 to this issue from https://github.com/pyodide/pyodide/issues/5013. If configured correctly, Pyodide would allow runtime for a Python interface for this WASM library through an Emscripten-compiled Python extension module.

Mizux: ok, I guess I understand what you want to do:
0. This is not something we support
1. currently our wrapper is generated by swig which is currently use to build 64bit code -> don't know how it work with WASM32 ptr size etc...
2. IIRC emcmake is working by injecting a CMake cross toolchain thus my bet is the Toolchainfile do not provide a python thus the issue.
3. I would suggest to play with my pet project: https://github.com/Mizux/python-native which should have the same cmake stuff than or-tools so once you have something work there it should be easier to patch/fixup or-tools and dev iteration should be faster IMHO...

sorry this comment is little bit dry, will try to reformulate if you want...

PR #2407, opened 2021-02-21
Add special error message for .cpp SOURCE

hevrard:
Before this change, when a user of an or-tools release tries to use a
".cpp" file with e.g.:

    make run SOURCE=./foo.cpp

The error message is:

    make: *** No rule to make target 'run'.  Stop.

This message does not really help to understand that the makefile is
designed to recognize only ".cc" files as C++ sources.

This change adds a specific error message when the source is a ".cpp"
file, inviting the user to use ".cc" instead.

### DIFF
--- a/tools/Makefile.cc.java.dotnet
+++ b/tools/Makefile.cc.java.dotnet
@@ -304,6 +304,15 @@ run: build
 	$(BIN_DIR)$S$(SOURCE_NAME)$E $(ARGS)
 endif # ifeq ($(SOURCE_SUFFIX),.cc)
 
+# Special error message for the common case of using ".cpp" instead of ".cc"
+ifeq ($(SOURCE_SUFFIX),.cpp)
+.PHONY: build
+build:
+	@echo Error: SOURCE has a .cpp extension, for C++ please use .cc
+.PHONY: run
+run: build
+endif # ifeq ($(SOURCE_SUFFIX),.cpp)
+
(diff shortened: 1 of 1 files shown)


hevrard:
Note to reviewers: I created this since I ran into that issue while following the C++ get started doc (which does say to use a ".cc" extension, I used ".cpp" out of habit). I tried to edit the Makefile that seems to be the one used as the top-level Makefile in release packages.

If you consider this too much of a corner case, I would understand, so no worries if you prefer to decline this PR.

Mizux:
Thanks for this PR, but since .cpp is pretty common IMHO i'll try to see if I can't modify it to support .cpp extension too
i.e. replacing https://github.com/google/or-tools/blob/b77bd3ac69b7f3bb02f55b7bab6cbb4bab3917f2/tools/Makefile.cc.java.dotnet#L287
by something like this
```Makefile
ifeq $(or ($(SOURCE_SUFFIX),.cc), ($(SOURCE_SUFFIX),.cpp))
```
note: not tested, will try it tomorrow

hevrard:
> Thanks for this PR, but since .cpp is pretty common IMHO i'll try to see if I can't modify it to support .cpp extension too

I thought about that too, I wasn't sure which approach best fit the project: enforcing ".cc" may help with keeping all files having the same extension -- but this may not be relevant for this specific Makefile.

PR #2558, opened 2021-05-18
Align C# ShiftSchedulingSat example to Python version

stuarthillary:
Changes in changesets 4ececbe and ab810ee were not applied to C# version

### DIFF
--- a/examples/dotnet/ShiftSchedulingSat.cs
+++ b/examples/dotnet/ShiftSchedulingSat.cs
@@ -390,7 +390,7 @@ public class ShiftSchedulingSat
         // Forbid sequences that are too short.
         foreach (var length in Range(1, hardMin))
         {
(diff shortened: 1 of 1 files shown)


lperron:
Thanks

Mizux (commented):
LGTM

Mizux:
ERRATUM: Seems the Python sample has been reverted to a "broken" version by
https://github.com/google/or-tools/commit/4ececbe448e1fe71a1e7e633414aedb3910a502e#diff-d464978be1d3fbcc3d70e907b2e2759615318df9e97f70f39db03c69e0d6a2a1
So
* .Net sample was OK so this patch should be reverted ^^;
* Python sample is broken and need to be fixed ASAP

BTW Many thanks to @stuarthillary for spotting the regression and the inconsistency !

PR #2587, opened 2021-06-08
fix #2575: Loader may not load native lib correctly due to resource n…

### LINKED ISSUE #2575: Native resource loader should avoid loading the same resource path from other jars.
Version: v9.0.
Language: Java

JVM could load ortools lib correctly.

**What did you see instead?**

The ortools lib could not be loaded correctly.

The current native resource Loader (for Java) firstly `getNativeResourceURI()` from the class path. However, it just locates the native lib via a resource path: Platform.RESOURCE_PREFIX + "/", for eg, darwin-x86-64/, or linux-x86-64.

It is possible that other jars has the same resource directory. In this case, the loader could not load the native lib sometimes.

xianyinxin:
…ame conflicting.


xianyinxin:
@Mizux Could you pls take a look?

xianyinxin:
@googlebot I signed it!

Mizux (approved):
LGTM, need to test...

stradivari96:
Small reminder to change merge target to `master`

xianyinxin:
@Mizux @stradivari96 I've changed the target branch to master, could you pls retrigger the ckecks if needed?

Issue #1435, opened 2019-07-20
Extend OR-Tools CMake Build Instructions


provinzio:
It took me quite a while to figure out how I can include OR-Tools to my c++ project using CMake.

Using the code below fixed my problem finally (described earlier [here](https://github.com/google/or-tools/issues/1236#issuecomment-493370811)).

It might be worth to add a hint to the [build instructions page](https://github.com/google/or-tools/blob/stable/cmake/README.md), if this isn't considered a bug.

```
list(APPEND CMAKE_PREFIX_PATH "<>/or-tools/build/dependencies/install/")

set(GFLAGS_USE_TARGET_NAMESPACE TRUE)
find_package(ortools CONFIG REQUIRED)

add_executable(project main.cpp)
target_link_libraries(project ortools::ortools)
```

lperron:
Not sure,

You can build or-tools using cmake and not make.
In that case, the dependencies will be built, but not in
dependencies/install/.
Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00

provinzio:
I had to build it with `cmake -DBUILD_DEPS:BOOL=ON -H. -Bbuild -G "Unix Makefiles"`.

Only using `cmake -H. -Bbuild -G "Unix Makefiles` gave me missing dependency errors.

lperron:
I am not the expert at all, but I believe you should do make install. This will package everything in one place.

provinzio:
I followed the build instruction page ending with `cmake --build build --target install`. I am still not able to find the dependencies anywhere in /usr/local. They only are in `/dependencies/install/`.

lperron:
OK. Unfortunately, Corentin is unavailable until early September.
Laurent Perron | Operations Research | lperron@google.com | (33) 1 42 68 53
00

apmanol:
with `BUILD_DEPS=ON` the  
`cmake/ortools/ortoolsTargets.cmake` created after the installation has the following entry:
```
set_target_properties(ortools::ortools PROPERTIES
  COMPATIBLE_INTERFACE_STRING "ortools_MAJOR_VERSION"
  INTERFACE_COMPILE_DEFINITIONS "USE_BOP;USE_GLOP;USE_CBC;USE_CLP"
  INTERFACE_COMPILE_FEATURES "cxx_std_11;cxx_std_11"
[...]

PR #2668, opened 2021-07-16
Cmake tweaks

jjacobsohn:
Provide example cmake commands to build the project and dependencies and place the build files and build products in a new subdirectory, as is recommended by the maintainers.

Also tweak the .gitignore file to avoid the last few spurious diffs caused by cmake, and re-organize some cmake related ignores from the IDE section.

Proposed as progress towards answering #1435

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@ bazel-*
 tools/docker/export
 
 # IDE stuff
+CMakeLists.txt.user
 *.userprefs
 .idea/*
 .idea/
@@ -88,7 +89,6 @@ examples/dotnet/obj
 .cmake/
 CMakeCache.txt
 CMakeFiles
-CMakeLists.txt.user
 DartConfiguration.tcl
 *build*/*
 build/
--- a/.gitignore
+++ b/.gitignore
@@ -57,12 +57,9 @@ bazel-*
 tools/docker/export
 
 # IDE stuff
-CMakeLists.txt.user
 *.userprefs
-*build*/*
 .idea/*
 .idea/
-build/
 cache/
 **/.vscode/*
 .DS_Store
@@ -87,3 +84,11 @@ examples/contrib/obj
 examples/dotnet/bin
 examples/dotnet/obj
 
+# cmake code gen (assumes -Bbuild)
+.cmake/
+CMakeCache.txt
+CMakeFiles
+CMakeLists.txt.user
+DartConfiguration.tcl
+*build*/*
+build/
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -103,12 +103,11 @@ For all of these options and parameters you have to use `-D<Parameter_name>=<val
 
 For example, to generate build files including dependencies in a new subdirectory called 'build', run:
 ```sh
-cmake -DBUILD_DEPS:BOOL=ON -H. -Bbuild
+cmake -S. -Bbuild  -DBUILD_DEPS:BOOL=ON
 ```
 and then build with:
 ```sh
-cd build
-cmake --build .
+cmake --build build
 ```
 
 Following is a list of available options, for the full list run:
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -100,7 +100,18 @@ test it on public CI and support can be broken.**
 
 There are several options that can be passed to CMake to modify how the code is built.<br>
 For all of these options and parameters you have to use `-D<Parameter_name>=<value>`.<br>
-Following a list of available options, for the full list run:
+
+For example, to generate build files including dependencies in a new subdirectory called 'build', run:
+```sh
+cmake -DBUILD_DEPS:BOOL=ON -H. -Bbuild
+```
+and then build with:
+```sh
+cd build
+cmake --build .
+```
+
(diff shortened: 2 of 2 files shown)


cmake/README.md
```diff
@@ -100,7 +100,18 @@ test it on public CI and support can be broken.**
 
 There are several options that can be passed to CMake to modify how the code is built.<br>
 For all of these options and parameters you have to use `-D<Parameter_name>=<value>`.<br>
-Following a list of available options, for the full list run:
+
+For example, to generate build files including dependencies in a new subdirectory called 'build', run:
+```sh
+cmake -DBUILD_DEPS:BOOL=ON -H. -Bbuild
```
Mizux, line 106: I would prefer to use:
```sh
cmake -S. -Bbuild  -DBUILD_DEPS:BOOL=ON 
```
note: `-H` and `-B` were hidden options (i.e. never officially supported) so we shouldn't depend on it, while `-S` and `-B` were officially publicly provided since CMake 3.13 (and or-tools require at least CMake 3.14 IIRC)
src:  https://cmake.org/cmake/help/latest/release/3.13.html#command-line

cmake/README.md
```diff
@@ -100,7 +100,18 @@ test it on public CI and support can be broken.**
...
+```
+and then build with:
+```sh
+cd build
```
Mizux, line 110: prefer to use `cmake --build build`

.gitignore
```diff
@@ -87,3 +84,11 @@ examples/contrib/obj
 examples/dotnet/bin
 examples/dotnet/obj
 
+# cmake code gen (assumes -Bbuild)
+.cmake/
+CMakeCache.txt
+CMakeFiles
+CMakeLists.txt.user
```
Mizux, line 91: technically `CMakeLists.txt.user` is generated by qtcreator aka an IDE and not by cmake (e.g. calling cmake configure/build from using a terminal)

.gitignore
```diff
@@ -87,3 +84,11 @@ examples/contrib/obj
...
+DartConfiguration.tcl
```
Mizux, line 92: are you sure of this DartConfiguration.tcl ? never saw it, out of curiosity which IDE did you use ?

Mizux (changes requested):

.gitignore
jjacobsohn: this appears to be an artifact generated by cmake to support ctest: https://gitlab.kitware.com/cmake/community/-/wikis/doc/ctest/Generating-Testing-Files 

I am using VisualStudio Code connected to a remote checkout on a Debian system. Pretty sure that VS is not involved but will dig a little deeper.

.gitignore
jjacobsohn: OK, thanks for straightening me out. Moving that file was an over-reach based purely on its name!

cmake/README.md
jjacobsohn: much cleaner, thanks

cmake/README.md
jjacobsohn: Helps to actually know vs copying things from StackOverflow! Thanks.

jjacobsohn:
Apologies for being unfamiliar with your review workflow. I added a commit to my fork of the or-tools repo, but don't know how to adjust this pull request to reference it (commit a07fac8dd6db4999912315906fd090ffbf1322ad).

.gitignore
jjacobsohn: I did a clean clone of or-tools on my Debian system from command line without VisualStudio running. The DartConfiguration.tcl file did not appear after 'cmake -S. -Bbuild  -DBUILD_DEPS:BOOL=ON', but was present in the build subdirectory after 'cmake --build build'. VS is exonerated!

Curiously the -enable_testing() statement which the CTest docs describe for creating the test build targets is removed from CMakeLists.txt by either scip-7.0.1.patch, and/or ZLIB.patch. It looks like this is a new symptom of an ongoing campaign to eliminate CTest!

I have a Dart.cmake template file in the directory where I checked out cmake 3.21 for local build, since Debian comes with a version that is too old to meet the or-tools minimum requirement. My theory is that my locally built cmake is inserting the .tcl file into the build, while your version/configuration of cmake is not.

I submit that including DartConfiguration.tcl in the cmake section of .gitignore can be useful and does no apparent harm.

jjacobsohn: Me again. Wrong about exonerating VS.

Opened VS Code, and now I have a new copy of DartConfiguration.tcl in the root of or-tools, along with the one that was created in .build by the command line cmake.

So the .gitignore entry is **not needed** for the command line one (it is already eliminated by *build*/*), but **is needed** if VisualStudio touches CMake (mine does because of installed extensions).

Which makes an argument to move DartConfiguration.tcl to the IDE section. Although it still is a CMake artifact. @Mizux what do you prefer?

Mizux: Since DartConfiguration.tcl is both a file generated by cmake and a file touch by vscode (CMake extension ?) let's keep it here (aka CMake group take precedence) -> I'll merge this PR as it

Mizux (approved):

PR #2786, opened 2021-09-23
Updating `java.cmake`

infvie:
For mac users, these environment vars will not be set by default when installing through `brew`.


infvie:
@Mizux please feel free to let me know if you need any help getting this issue resolved!

Issue #2621, opened 2021-06-30
springboot jar 形式启动


wu603772498:
java.lang.RuntimeException: java.nio.file.NoSuchFileException: /BOOT-INF/lib/ortools-linux-x86-64-9.0.9048.jar!/linux-x86-64/

Mizux:
maybe related to #2330

Robot-1x:
I think it is related to #2428. Maybe method works for you. 
I test in my Mac and Linux(centos 8.x) system. This method works well except M1 platform. :joy:
Besides, it is not fixed for jar file. I will provide a simple test project to describe this issue.

lperron:
Solution found here: https://github.com/google/or-tools/discussions/3553

Please reopen if it does not work.

PR #2831, opened 2021-10-10
More reliable native loading

magneticflux-:
This finally resolves some lingering [compatibility issues](https://github.com/magneticflux-/ortools-java#complex-classpaths-spring-etc) with alternate classloaders that my fork handles correctly. Once this is merged, I will deprecate my fork since all functionality will be included here.

It also removes the dependency on JNA, opting instead to simply check the results of `System.mapLibraryName`.


magneticflux-:
I see, however currently (as of 9.1) Java distribution is limited to the three cases I implemented. Additionally, the current solution is still limited by JNA support: https://github.com/java-native-access/jna/issues/1309.

I will see how JNA and LWJGL bootstrap their native libraries and try to emulate that, since they seem to be the best-supported/most popular Java libraries that bundle native code.

magneticflux-:
May fix https://github.com/google/or-tools/issues/2621

ortools/java/com/google/ortools/Loader.java
```diff
@@ -13,102 +13,75 @@
...
+  private static final String RESOURCE_SUFFIX;
+  private static final String[] FILES_TO_EXTRACT;
+  static {
+    RESOURCE_SUFFIX = System.mapLibraryName("test").split("test\\.")[1];
+    switch (RESOURCE_SUFFIX) {
+      case "dll":
+        RESOURCE_PREFIX = "win32-x86-64";
+        FILES_TO_EXTRACT = new String[]{"jniortools"};
+        break;
+      case "so":
```
Mizux, line 37: this won't work since now we support `linux-86-64` and `linux-arm64` you can't rely on the dynamic library extension

ortools/java/com/google/ortools/Loader.java
```diff
@@ -13,102 +13,75 @@
...
+        RESOURCE_PREFIX = "linux-x86-64";
+        FILES_TO_EXTRACT = new String[]{"libjniortools", "libortools"};
+        break;
+      case "dylib":
```
Mizux, line 41: this won't work since now we support `darwin-86-64` and `darwin-arm64` (Apple M1) -> you can't rely on the dynamic library extension...

lperron:
Most likely, this will never been merged.

PR #3016, opened 2021-12-13
Add astar.cc to shortestpaths target

kgreenek:
Without this, using the AStarShortestPath function from
//ortools/graph/shortestpaths.h results in a linker error because the
function definition is not included in the build.

I also fixed some reorder compiler warnings in astar.cc.


Mizux (approved):
LGTM

PR #3027, opened 2021-12-20
Replace deprecated SCIP cons_quadratic.h

jediahkatz:
The function `SCIPcreateConsBasicQuadratic()` along with the entire `cons_quadratic.h` file are deprecated -- instead, we should use `SCIPcreateConsBasicQuadraticNonlinear()` which takes the same arguments. Right now, attempting to use quadratic constraints or objectives results in all variables taking a zero value.

For reference see the SCIP docs:
SCIPcreateConsBasicQuadratic: https://www.scipopt.org/doc/html/group__CONSHDLRS.php#gad3707e7f7166bea83b7713cf2e52b0db
SCIPcreateConsBasicQuadraticNonlinear: https://www.scipopt.org/doc/html/group__CONSHDLRS.php#gabef6770f404754e7503f68a7da9ad6da

### DIFF
--- a/ortools/linear_solver/scip_proto_solver.cc
+++ b/ortools/linear_solver/scip_proto_solver.cc
@@ -42,7 +42,7 @@
 #include "ortools/util/lazy_mutable_copy.h"
 #include "scip/cons_disjunction.h"
 #include "scip/cons_linear.h"
-#include "scip/cons_quadratic.h"
+#include "scip/cons_nonlinear.h"
 #include "scip/pub_var.h"
 #include "scip/scip.h"
 #include "scip/scip_param.h"
@@ -240,7 +240,7 @@ absl::Status AddQuadraticConstraint(
   }
 
   RETURN_IF_SCIP_ERROR(
-      SCIPcreateConsBasicQuadratic(scip,
+      SCIPcreateConsBasicQuadraticNonlinear(scip,
                                    /*cons=*/scip_cst,
                                    /*name=*/gen_cst.name().c_str(),
                                    /*nlinvars=*/lsize,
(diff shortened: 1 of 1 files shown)


Mizux:
This break the bawel build, maybe need to bump the version first:
https://github.com/google/or-tools/blob/2cb85b4eead4c38e1c54b48044f92087cf165bce/WORKSPACE#L84-L90

jediahkatz:
Hmm, strange because `nonlinear.h` exists in commit `6acb7222e1b871041445bee75fc05bd1bcaed089` of the SCIP repo. Unfortunately though, I realized that the issue is less simple than I originally thought. The all-zero values problem was caused by me using `Solve()` instead of `SolveWithProto()`, but when I use `SolveWithProto()` then I get an error with the following trace:
```
Traceback (most recent call last):
  File "ortools_samples/quadratic.py", line 88, in <module>
    print(pywraplp.Solver.SolveWithProto(linear_solver_pb2.MPModelRequest(model=input_proto.model), solution))
  File "/home/jediahkatz/.cache/pypoetry/virtualenvs/pythonsat-yDcEWZll-py3.8/lib/python3.8/site-packages/ortools/linear_solver/pywraplp.py", line 358, in SolveWithProto
    return _pywraplp.Solver_SolveWithProto(model_request, response)
TypeError: in method 'Solver_SolveWithProto', argument 1 of type 'operations_research::MPModelRequest const &'
```
It seems to be a Python/SWIG issue -- the C++ function expects `operations_research::MPModelRequest const &`, and I think the Python `MPModelRequest` object is not properly passed as `const`. For now I have given up on using the quadratics with ortools but maybe at some point I will take another look (but feel free to close the PR if you don't care about such a niche bug).

lperron:
The linear_solver.i is missing the wrapping of the MPModelRequest.
I will fix this asap.
00



Le dim. 2 janv. 2022 à 02:17, Jediah Katz ***@***.***> a
écrit :

lperron:
And pushed to master.
00

jediahkatz:
Ah, thanks so much. I'm not super familiar with how SWIG works but I'll check out your commit for curiosity's sake. Bonne année!

PR #3046, opened 2021-12-29
updated comment at requests for shift_scheduling_sat.py

georgiossalon:

### DIFF
--- a/examples/python/shift_scheduling_sat.py
+++ b/examples/python/shift_scheduling_sat.py
@@ -215,7 +215,7 @@ def solve_shift_scheduling(params, output_proto):
     # Request: (employee, shift, day, weight)
     # A negative weight indicates that the employee desire this assignment.
     requests = [
-        # Employee 3 wants the first shift of Saturday off.
+        # Employee 3 does not want the first shift on the first Saturday (negative weight).
         (3, 0, 5, -2),
         # Employee 5 does not want a night shift on the second Thursday (negative weight).
         (5, 3, 10, -2),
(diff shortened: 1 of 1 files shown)


examples/python/shift_scheduling_sat.py
```diff
@@ -215,11 +215,11 @@ def solve_shift_scheduling(params, output_proto):
     # Request: (employee, shift, day, weight)
     # A negative weight indicates that the employee desire this assignment.
     requests = [
-        # Employee 3 wants the first Saturday off.
+        # Employee 3 wants the first shift of Saturday off.
```
Mizux, line 218: I would rephrase it too to be consistency:
"Employee 3 does not want the first shift on the first Saturday (negative weight)."

lperron: and please against master

georgiossalon: @lperron The pull request goes into google:master. Does this fulfill your request?

Mizux: I guess laurent was trick by the name of your branch stable otherwise everything is good !
thanks for the PR

lperron:
I will revert this. The direction is wrong. value > 0 means the shift is not desired.
value < 0 means the shift is desired.

georgiossalon:
I am sorry about that. Then line 218 (Saturday off) should get corrected.

Mizux:
fixed by https://github.com/google/or-tools/commit/6f24d2a0eaa7176cd0ff783bcff4d8c8139447f4

Issue #3060, opened 2022-01-04
Bazel target //ortools/base:recordio relies on system-installed zlib


kgreenek:
**What version of OR-Tools and what language are you using?**
Version: master
Language: C++

**What operating system (Linux, Windows, ...) and version?**
Linux Ubuntu 20.04

**What did you do?**
Build with bazel and deploy to a system without zlib installed.
In my case, I am using the `//ortools/graph:shortestpaths` library.
I am using [rules_docker](https://github.com/bazelbuild/rules_docker) to create an image for my cc_binary.
When I run the container, I get a runtime error, due to the reliance on z.so not being available at runtime.

**What did you expect to see**
Bazel-built libraries (if possible) should not rely on system-installed libraries.

**What did you see instead?**
Runtime error that z.so is not found.

**Anything else we should know about your project / environment**
It looks like `@zlib` is already a WORKSPACE dependency for windows, but on linux it uses the system library instead of the bazel-built one.

lperron:
Why don't you install zlib in your docker?

kgreenek:
I'm using rules_docker with the cc_image rule.

kgreenek:
I don't know the full context for why the system-installed version is used here on linux. But if zlib is managed by bazel, then I don't have to use docker to package it -- it'll be statically linked and I can just package my binary. The cc_image rule is very cool in that it is distroless. It only has the c++ runtime and nothing else, so it is very tiny. But that does mean you can't just system install things easily. You  need to use bazel to manage your dependencies in order to use it out of the box.

PR #3062, opened 2022-01-05
Use bazel workspace target for zlib

kgreenek:
Fixes https://github.com/google/or-tools/issues/3060

Without this, linux builds will fail at runtime if z.so is not
installed on the deploy platform. This is problematic for example, when
using rules_docker to create a minimal image for a cc_binary to run.

NOTE: Users who want to use a system-installed version of zlib can still
do so by creating a library in their workspace called `@zlib:zlib` with
`-lz` in its linkopts. This gives flexibility at the workspace level for
choosing how to build zlib.

### DIFF
--- a/ortools/base/BUILD.bazel
+++ b/ortools/base/BUILD.bazel
@@ -184,24 +184,14 @@ cc_library(
     hdrs = [
         "recordio.h",
     ],
-    linkopts = select({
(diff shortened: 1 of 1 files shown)


Mizux:
Nice catch thanks !

PR #3068, opened 2022-01-11
Fix CMake package config file

gmgunter:
There are currently a few issues with how the COIN-OR dependencies are resolved when using `find_package(ortools CONFIG)` to add or-tools to an external CMake project.

* With CMake >=3.9.6, we try to find the Clp and Cbc packages using the CONFIG mode of `find_package`. This always fails since neither [coin-or/Clp](https://github.com/coin-or/Clp) nor [coin-or/Cbc](https://github.com/coin-or/Cbc) provides a CMake config file.

* If we address the above issue by instead using the MODULE mode of `find_package`, then we need to add `FindCbc.cmake` and `FindCpl.cmake` scripts to the CMAKE_MODULE_PATH in order to teach CMake how to find these dependencies. I propose that or-tools should install these scripts alongside its config file so that they're available to external CMake projects.

* Finally, the `FindCbc.cmake` script included with or-tools defines the variable `CBC_FOUND` when it's successful, rather than `Cbc_FOUND` as expected ([CMake variables are case-sensitive](https://stackoverflow.com/a/35034125)). The `FindCpl.cmake` script has a similar issue. As a result, even when the above two points are addressed, `find_package(ortools CONFIG)` will still fail because CMake erroneously thinks that these two dependencies weren't succesfully found.

This PR attempts to address the above issues with the following changes:

* The CMake config file is modified to allow searching for Cbc and Cpl using the MODULE mode of `find_package`.

* `FindCbc.cmake` and `FindCpl.cmake` are installed with the CMake package config files and are added to the CMAKE_MODULE_PATH if `USE_COINOR` was truthy.

* The `FindCbc.cmake` and `FindCpl.cmake` files are modified to change the case of the variables they export.

### DIFF
--- a/cmake/cpp.cmake
+++ b/cmake/cpp.cmake
@@ -279,6 +279,12 @@ install(
   "${PROJECT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
   DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}"
   COMPONENT Devel)
+install(
+  FILES
+  "${PROJECT_SOURCE_DIR}/cmake/FindCbc.cmake"
+  "${PROJECT_SOURCE_DIR}/cmake/FindClp.cmake"
+  DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}/modules"
+  COMPONENT Devel)
 
 # add_cxx_sample()
 # CMake function to generate and build C++ sample.
--- a/cmake/cpp.cmake
+++ b/cmake/cpp.cmake
@@ -279,12 +279,14 @@ install(
   "${PROJECT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
   DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}"
   COMPONENT Devel)
-install(
-  FILES
-  "${PROJECT_SOURCE_DIR}/cmake/FindCbc.cmake"
-  "${PROJECT_SOURCE_DIR}/cmake/FindClp.cmake"
-  DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}/modules"
-  COMPONENT Devel)
+if(USE_COINOR)
+  install(
+    FILES
+    "${PROJECT_SOURCE_DIR}/cmake/FindCbc.cmake"
+    "${PROJECT_SOURCE_DIR}/cmake/FindClp.cmake"
+    DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}/modules"
+    COMPONENT Devel)
+endif()
 
 # add_cxx_sample()
 # CMake function to generate and build C++ sample.
--- a/cmake/FindCbc.cmake
+++ b/cmake/FindCbc.cmake
@@ -22,9 +22,9 @@ Cbc_FOUND          - True if Cbc found.
 #]=======================================================================]
 find_package(PkgConfig REQUIRED)
 
-pkg_check_modules(CBC REQUIRED cbc IMPORTED_TARGET GLOBAL)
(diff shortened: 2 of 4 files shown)


Mizux:
1) I've fork coin-or few years ago to add CMake support, they never want to integrate it...

gmgunter:
>I've forked coin-or repos (Cbc, Clp, Cgl, Osi, CoinUtils) few years ago to add CMake support, they never want to integrate it...

That's too bad. The most elegant solution would definitely be for these upstream packages to add CMake support.

>Need a job to test the CMake based build of or-tools against coin-or installed system-wide

This would be really nice. I installed or-tools via the [conda-forge package](https://anaconda.org/conda-forge/ortools-cpp), which builds with [`BUILD_DEPS=OFF`](https://github.com/conda-forge/or-tools-feedstock/blob/c331915da5c25ecabb4d73633d503f7427c5cf1e/recipe/build-cpp.sh#L7).

cmake/cpp.cmake
```diff
@@ -279,6 +279,12 @@ install(
   "${PROJECT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
   DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}"
   COMPONENT Devel)
+install(
```
Mizux, line 282: can you guard it with a `if(USE_COINOR)`, i.e. if the coin-or support is disable there is no interest to install module

Mizux (changes requested):

Mizux (approved):

Mizux:
FYI I'll need to change the module CMake since:
1. It does not support super build i.e. if Cbc target already present the module should be a noop
2. It does not support cbc_config.cmake file (like we have when building cbc or using vcpkg) so first module should try a to perform a find_package( CONFIG) then if found -> nothing to do, otherwise try to locate it using the pkg plugin...

On my way to fix it, if you have some regression in your worflow, please let me know...

gmgunter:
Sounds good :ok_hand:

Neumann-A:
@Mizux I'll probably be the one to rip out all the `CMakeLists.txt` vcpkg vendors. Including any exported targets which have been added and are currently in disagreement with the maintainer guidelines of vcpkg. Either you get upstream to accept your CMake versions or pkg-config will be the preferred way to consume coin-or libraries.

Mizux:
@Neumann-A
1. Do you have a link to this maintainer guidelines assertion ?
2. Does pkg-config build will work on Windows and nicely integrate with vcpkg ?
3. How to deal (nicely) with a CMake toolchain file if using pkg-config/autotools based build ?
4. How about having a Community supported CMake fork ?

note: here pkg-config/autotools based build means what coin-or provide if I understand correctly your point
note2: IIRC it was a user of openframework who provide a first ugly CMakeLists.txt patch on vcpkg

Neumann-A:
> Do you have a link to this maintainer guidelines assertion ?

a) The current name of most coin-or ports in vcpkg is wrong according to https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#check-names-against-other-repositories (basically the ports need a prefix `coin-or-*`

b) `unofficial::` namespace: https://github.com/microsoft/vcpkg/blob/master/docs/maintainers/maintainer-guide.md#add-cmake-exports-in-an-unofficial--namespace

> Does pkg-config build will work on Windows and nicely integrate with vcpkg ?
> How to deal (nicely) with a CMake toolchain file if using pkg-config/autotools based build ?

vcpkg has a port for pkgconf. If it is installed CMake's`FindPkgConfig` will automatically start working. be aware that it breaks multi-config usage (e.g. no interactive switching of the build type, as long as you stay with one build type you are fine). For ports internal to vcpkg it will start to automatically work due to how the `vcpkg_` function are setup and if additional downstream usage agrees with upstream provided usage. 

> How about having a Community supported CMake fork ?

No. Upstream it or leave it. Source packages have to come from the authoritative source. 

So I already started ripping it out: https://github.com/microsoft/vcpkg/pull/29398. basically the only thing left to do is fixing openmvg and getting the linux VM to have autoconf 2.71

PR #3081, opened 2022-01-14
improve allocation on some tests

bollhals:
Reduces the allocation of some tests, so it's easier to spot where the remaining allocations are.
No functional changes

Changes:
- use ValueTuple instead of Tuple
- Reuse Lists
- Set Capacity on some large proto model


lperron:
much appreciated.

Thanks

PR #3085, opened 2022-01-15
set capacity before adding if size is known

bollhals:
Sets the capacity of Lists and RepeatedField if it is known
=> Avoids allocation of intermediate arrays (when a list is growing and growing) or too large arrays (e.g. protobuf allocates 8-size arrays per default, which can depending on the situation be way too large)

|                     Method |     Mean |     Error |    StdDev |   Gen 0 |  Gen 1 | Allocated |
|--------------------------- |---------:|----------:|----------:|--------:|-------:|----------:|
| NursesSat_WithOutputToNull_pr | 1.029 ms | 0.0100 ms | 0.0094 ms | 11.7188 | 5.8594 |     58 KB |
| NursesSat_WithOutputToNull_master | 1.046 ms | 0.0068 ms | 0.0063 ms | 11.7188 | 5.8594 |     62 KB |

=> 1.6 % reduction of duration
=> 6.5 % managed allocation reduction


lperron:
Thanks

PR #3086, opened 2022-01-17
improve Value function and introduce a simplified one for IntVar

bollhals:
This pullrequest targets the Value functions by
- introducing a Value(IntVar) overload to simplify
- simplify the Value(LinearExpr)

|                     Method |     Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Allocated |
|--------------------------- |---------:|----------:|----------:|-------:|-------:|----------:|
| NursesSat_WithOutputToNull_pr | 1.039 ms | 0.0031 ms | 0.0029 ms | 7.8125 | 3.9063 |     37 KB |
| NursesSat_WithOutputToNull_master | 1.072 ms | 0.0208 ms | 0.0185 ms | 11.7188 | 5.8594 |     58 KB |

=> Slight reduction of duration but a massive reduction (36%) of managed allocation

### DIFF
--- a/ortools/sat/csharp/CpSolver.cs
+++ b/ortools/sat/csharp/CpSolver.cs
@@ -14,6 +14,7 @@
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using System.Threading;
 
 namespace Google.OrTools.Sat
 {
@@ -28,6 +29,20 @@ namespace Google.OrTools.Sat
  */
 public class CpSolver
 {
+    private static Queue<Term> _cachedQueue;
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static Queue<Term> RentQueue()
+    {
+        return Interlocked.Exchange(ref _cachedQueue, null) ?? new Queue<Term>(2);
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static void ReturnQueue(Queue<Term> queue)
+    {
+        Interlocked.Exchange(ref _cachedQueue, queue);
+    }
+
     /** <summary>Solves the given model, and returns the solve status.</summary> */
     public CpSolverStatus Solve(CpModel model, SolutionCallback cb = null)
     {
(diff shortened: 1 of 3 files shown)


ortools/sat/csharp/CpSolver.cs
```diff
@@ -14,6 +14,7 @@
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using System.Threading;
```
lperron, line 17: remove

ortools/sat/csharp/CpSolver.cs
```diff
@@ -28,6 +29,20 @@ namespace Google.OrTools.Sat
  */
 public class CpSolver
 {
```
lperron, line 31: just private Queue<Term> queue_;

ortools/sat/csharp/CpSolver.cs
```diff
@@ -28,6 +29,20 @@ namespace Google.OrTools.Sat
...
+    private static Queue<Term> _cachedQueue;
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
lperron, line 34: no need for this. Value() is called after the search. We can document it is not MT-HOT, or make it synchronied.

ortools/sat/csharp/CpSolver.cs
```diff
@@ -148,56 +163,72 @@ public CpSolverResponse Response
         }
     }
 
+    /**
+     * <summary>
+     * Returns the value of an integer variable in the last solution found.
+     * </summary>
+     */
+    public long Value(IntVar intVar)
```
lperron, line 171: Nice!

ortools/sat/csharp/CpSolver.cs
```diff
@@ -148,56 +163,72 @@ public CpSolverResponse Response
...
      * </summary>
      */
     public long Value(LinearExpr e)
     {
-        List<Term> terms = new List<Term>();
-        terms.Add(new Term(e, 1));
         long constant = 0;
+        long coefficient = 1;
+        LinearExpr expr = e;
+        Queue<Term> terms = RentQueue();
```
lperron, line 188: just use the field in the CpSolver class.

ortools/sat/csharp/IntegerExpressions.cs
```diff
@@ -69,6 +69,21 @@ public static void AddOrIncrement(this Dictionary<IntVar, long> dict, IntVar key
             list.Capacity = Math.Max(list.Count + collection.Count, list.Capacity);
         }
     }
+
+#if NETFRAMEWORK
```
lperron, line 73: which version NETFRAMEWORK covers ? We will soon support only net6 and above.

ortools/sat/csharp/IntegerExpressions.cs
```diff
@@ -69,6 +69,21 @@ public static void AddOrIncrement(this Dictionary<IntVar, long> dict, IntVar key
...
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
lperron, line 74: Not sure adding a method is worth it if it is called only once.

ortools/sat/csharp/SearchHelpers.cs
```diff
@@ -31,50 +31,66 @@ public class CpSolverSolutionCallback : SolutionCallback
      */
     public long Value(LinearExpr e)
     {
-        List<Term> terms = new List<Term>();
-        terms.Add(new Term(e, 1));
         long constant = 0;
+        long coefficient = 1;
+        LinearExpr expr = e;
+        Queue<Term> terms = CpSolver.RentQueue();
```
lperron, line 37: Just add a Queue field in the callback. Only initialize it on demand. By construction, only one callback can be executed at a given time.

ortools/sat/csharp/CpSolver.cs
bollhals: I'll reply here but this is related to other comments as well.

The intent of the queue being static is twofold:
- easy access from multiple classes (CpSolver, Callback, LinearExpression (probably next PR)) (As non static would require met to pass it around)
- share the queue between different solver instances (No need to allocate in the default sequential case)

Making it synchronized works, but will most likely make it slower than the Interlocked.Exchange with the same benefits. (And is only needed if it stays static, if it is not, then there is no point in any protection, since the rest of the code here isn't protected against multithread access)

ortools/sat/csharp/IntegerExpressions.cs
bollhals: All of them technically, but since as far as I saw it's currently only built for .NET 4.5.2, Core 3.1 and .NET 6. So it is only used for the first one, the last 2 do have this API on the Queue class, so they don't need it. [See here](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.queue-1.trydequeue?view=net-6.0)

ortools/sat/csharp/IntegerExpressions.cs
bollhals: The plan would be to use it soon in another place, next PR, this is why I added it here.

lperron:
This is a big no
 Sorry. Nothing static, nothing shared between solvers.

lperron:
What is the sequential case for you?

lperron:
I have cherry picked some of the changes (queue based evaluation, Value(IntVar), TryDequeue.

Let's use that as a baseline against more complex design.

My take:
  - callbacks are rare and slow anyway. Their execution is protected by a c++ mutex. So a single non-protected queue seems a good idea to me. The extra allocation pales against the machinery it involves as a director swig object.
  - the solver is usually called in a sequential model, whatever the number of workers. So a single queue is a good design. The majority of the code should be using the IntVar API anyway, So the queue will most likely never be allocated.
  - I really think adding contention between solvers with a shared/static object is a bad idea.
  - For the linear expr GetVarValueMap, it is used only in CpModel. We can change the API to pass the queue to the GetVarValueMap, and have one queue always created in the CpModel class.

bollhals:
Thanks for the feedback, really appreciated, I'll check and back to you. 👍

lperron:
I will push the GetVarValueMap(..., terms) very soon

00

bollhals:
> I will push the GetVarValueMap(..., terms) very soon 

Now the public method does have the queue parameter, feels strange to need to supply one. => Is this mehtod even suppose to be public? If so would it make more sense to have an overload with the old syntax to prevent a breaking change and forwards to the other method with the queue of the instance field?

And does it make sense to have a version of it with an IntVar as well for this function? (Haven't checked the usage / benefits, just popped in my mind)

lperron:
The method is not meant to be used, but I do not see the need to restrict its access, nor provide a second version.

lperron:
I have reduced the visibility of internal methods to internal.

bollhals:
> * callbacks are rare and slow anyway. Their execution is protected by a c++ mutex. So a single non-protected queue seems a good idea to me. The extra allocation pales against the machinery it involves as a director swig object.
> * the solver is usually called in a sequential model, whatever the number of workers. So a single queue is a good design. The majority of the code should be using the IntVar API anyway, So the queue will most likely never be allocated.
> * I really think adding contention between solvers with a shared/static object is a bad idea.
> * For the linear expr GetVarValueMap, it is used only in CpModel. We can change the API to pass the queue to the GetVarValueMap, and have one queue always created in the CpModel class.

Fine by me, I think the most important part is that it doesn't allocate in every call. 

Just to highlight the difference, in the current way of master, it potentially allocates a queue in 3 places and nothing is shared between different instances. By using a static, you could reduce it to 1 created in total. Contention between solvers is not really existent, as with the interlocked class we try to get the cached queue and if not we just create a new one to use (which only can happen if two threads would need the queue at the same time, e.g. two CpSolver by different threads)

But as I said this is in the end only a very very minor optimization that as you've explained is unlikely to make it in.

Overall I think you've cherry picked the important parts of this PR, so I'm going to close this.

PR #3093, opened 2022-01-19
simplify reading from returned buffer

bollhals:
``This pr contains optimization around (de)serializing proto messages.

- avoid calling message.CalculateSize() twice
- instead of copying 4 bytes to array and read it as int, read directly from unmanaged memory
- instead of allocating the array to copy from unmanged, it is now using the arraypool
- using <messagetype>.Parser.ParseFrom instead of CodedInputStream and MergeFrom

Two outstanding questions:
- What about this comment 
> TODO(user): delete the C++ buffer.

Does that mean we're constantly "leaking" memory? Is this fixable?
- I was unable to optimize it further due to my lack of understanding of the code generation. Optimal would be in my opinion:
For methods with input/output
```cs
public Google.OrTools.Sat.CpSolverResponse Solve(Google.OrTools.Sat.CpModelProto model_proto) {
    System.IntPtr data = operations_research_satPINVOKE.SolveWrapper_Solve(swigCPtr, ProtoHelper.ProtoToByteArray(model_proto, out var buf), buf);
    int size = System.Runtime.InteropServices.Marshal.ReadInt32(data);
    if (size > buffer.Length) {
        System.Buffers.ArrayPool<byte>.Shared.Return(buf);
        buf = System.Buffers.ArrayPool<byte>.Shared.Rent(size);
    }
    data += sizeof(int);
    System.Runtime.InteropServices.Marshal.Copy(data, buf, 0, size);
    // TODO(user): delete the C++ buffer.
    try {
      return Google.OrTools.Sat.CpSolverResponse.Parser.ParseFrom(System.MemoryExtensions.AsSpan(buf, 0, size));
    } catch (Google.Protobuf.InvalidProtocolBufferException /*e*/) {
      throw new System.Exception("Unable to parse Google.OrTools.Sat.CpSolverResponse protocol message.");
    }
    finally {
      System.Buffers.ArrayPool<byte>.Shared.Return(buf);
    }
  }
```

### DIFF
--- a/ortools/util/csharp/ProtoHelper.cs
+++ b/ortools/util/csharp/ProtoHelper.cs
@@ -14,18 +14,16 @@
 namespace Google.OrTools
 {
 
-using System;
 using Google.Protobuf;
 
 public static class ProtoHelper
 {
-    public static byte[] ProtoToByteArray(IMessage message)
+    public static int ProtoToByteArray(IMessage message, out byte[] buffer)
     {
         int size = message.CalculateSize();
-        byte[] buffer = new byte[size];
-        CodedOutputStream output = new CodedOutputStream(buffer);
-        message.WriteTo(output);
-        return buffer;
+        buffer = new byte[size];
+        message.WriteTo(buffer);
+        return size;
     }
 }
 } // namespace Google.OrTools
--- a/ortools/util/csharp/proto.i
+++ b/ortools/util/csharp/proto.i
@@ -50,7 +50,7 @@
 %typemap(ctype)  PROTO_TYPE* INPUT, PROTO_TYPE& INPUT "int " #param_name "_size, uint8_t*"
 %typemap(imtype) PROTO_TYPE* INPUT, PROTO_TYPE& INPUT "int " #param_name "_size, byte[]"
 %typemap(cstype) PROTO_TYPE* INPUT, PROTO_TYPE& INPUT "CSharpProtoType"
-%typemap(csin)   PROTO_TYPE* INPUT, PROTO_TYPE& INPUT "$csinput.CalculateSize(), ProtoHelper.ProtoToByteArray($csinput)"
+%typemap(csin)   PROTO_TYPE* INPUT, PROTO_TYPE& INPUT "ProtoHelper.ProtoToByteArray($csinput, out var buffer), buffer"
 %typemap(in)     PROTO_TYPE* INPUT, PROTO_TYPE& INPUT {
   $1 = new CppProtoType;
   bool parsed_ok = $1->ParseFromArray($input, param_name ## _size);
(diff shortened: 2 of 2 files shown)


lperron:
About the comment, I am not sure. It should be checked with a repetitive solve() to see if we leak the response buffer each time. Maybe we need to use newfree. http://www.swig.org/Doc1.3/Typemaps.html#Typemaps_nn34

bollhals:
> * I was unable to optimize it further due to my lack of understanding of the code generation. Optimal would be in my opinion:
>   For methods with input/output
> 
> ```cs
>     int size = System.Runtime.InteropServices.Marshal.ReadInt32(data);
>     if (size > buffer.Length) {
>         System.Buffers.ArrayPool<byte>.Shared.Return(buf);
>         buf = System.Buffers.ArrayPool<byte>.Shared.Rent(size);
>     }

> ```

Any idea on how this could be achieved?
Either
- special block of code when the method has input and output
- additional block after the $imcall for when the method has an input

lperron:
I believe this is the responsibility of the Rent() part to be efficient. So I would not try to optimize it.
We could gain if we could use the C++ buffer directly in C#, as the memory is pinned anyway.

Not sure it is allowed though.

bollhals:
There are three things to discuss here:

**1. Avoid having to use a managed byte array when returning a value from c++**
To my knowledge, one option would be to enable the [unsafe switch](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/language#allowunsafeblocks) in the project, this makes it possible to create a span over the native memory to parse the protobuf model. I can set up a PR if the [unsafe switch](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/language#allowunsafeblocks) is fine for you.

**2. Pool byte array for input call**
Currently only the response is pooled, not the input. This is due to `%typemap(csin)   PROTO_TYPE* INPUT, PROTO_TYPE& INPUT "ProtoHelper.ProtoToByteArray($csinput, out var buffer), buffer"` only allow modifying the parameter, but due to needing to return the buffer, some code after the call is required. Question is how to achieve that

**3. Reusing buffer instead of return and rent a new one**
This is only an optimization. If the method has input and output, if we can add code for just this case, then we can implement the if case shown above.

bollhals:
> 1. I expect this to have consequences if the assembly is loaded on a server. So I guess no.

This doesn't change anything except allowing the use of the unsafe keyword, which allows creating a span<byte> from a byte pointer. (Thus not needing to use Marshal.Copy, nor the array itself to copy)

> 2. You can release the buffer after the ParseFromArray in the typemap(in). Did I miss something ?

If I add anything there, nothing changes. I can only modify the csin part, but that only adjust the parameters, but I can't add real code there. 🤔

lperron:
1) not convinced :-)

2) What code do you want to add ?

bollhals:
> 1. not convinced :-)

See #3097 


> 2. What code do you want to add ?

The Solve method for the CpModelProto would need the following:
```cs
public Google.OrTools.Sat.CpSolverResponse Solve(Google.OrTools.Sat.CpModelProto model_proto) {
  System.IntPtr dataPointer = operations_research_satPINVOKE.SolveWrapper_Solve(swigCPtr, ProtoHelper.ProtoToByteArray(model_proto, out var buffer), buffer);

  // +++++++++ This line would need to be added +++++++++
  System.Buffers.ArrayPool<byte>.Shared.Return(buffer);

  int size = System.Runtime.InteropServices.Marshal.ReadInt32(data);
  // rest of the method
}

public static int ProtoToByteArray(IMessage message, out byte[] buffer) {
  int size = message.CalculateSize();
  // +++++++++ The next line is replaced with the next after that +++++++++
  // buffer = new byte[size];
  buffer = System.Buffers.ArrayPool<byte>.Shared.Rent(size);
  message.WriteTo(buffer);
  return size;
}
```

This would enable the buffer to be pooled, thus avoiding the array allocation.

But I can't seem to find the place to add the line with the Return(buffer) to do this.

lperron:
I approved #3097, but I am pretty sure it will be rejected by some pedantic environment, and I will need to backtrack it.
We already have a lot of problems with assembly signature. So I expect some environment to reject assemblies with unsafe code

For 2, this only happens once at the start of the solve. Compared to the number of times we copy and scan the model, this is not worth spending more time on it.

bollhals:
> I approved #3097, but I am pretty sure it will be rejected by some pedantic environment, and I will need to backtrack it. We already have a lot of problems with assembly signature. So I expect some environment to reject assemblies with unsafe code

To my knowledge, there is no difference in the generated assembly by using unsafe code, hence it is not detectable.

PR #3095, opened 2022-01-20
Fix magic square model in C++

mpelleau:
Fix indexes for the second diagonal, and comment the loop for the alldifferent constraint


lperron:
Hi Marie, 

I have cherry-picked the loop on all-different.
We do not accept PR on the stable branch, only on the master branch. Unfortunately, the code has changed a lot since the last release.

I believe the index error is fixed on the new code.

PR #3100, opened 2022-01-25
improve GetVarValueMap

bollhals:
This  PR improves the GetVarValueMap function and the ones that uses it.

Mainly these changes were made:
- The dictionary used for returning the data is stored in the CpModel and reused between the calls. 
- Change the key type of the dictionary from IntVar to int, as its index was always used => This enables the usage of `AddRange(dict.Keys)` which is baster than iterating over it manually.
- the initial_coeff was always 1L, so I set it inside the function instead of a parameter

### DIFF
--- a/ortools/sat/csharp/CpModel.cs
+++ b/ortools/sat/csharp/CpModel.cs
@@ -1194,25 +1194,24 @@ public class CpModel
         var dict = var_value_map_;
         dict.Clear();
         long constant = LinearExpr.GetVarValueMap(expr, dict, terms_);
-        long mult = negate ? -1 : 1;
 
         LinearExpressionProto linear = new LinearExpressionProto();
         var dictCount = dict.Count;
         linear.Vars.Capacity = dictCount;
         linear.Vars.AddRange(dict.Keys);
         linear.Coeffs.Capacity = dictCount;
-        if (!negate)
-        {
-            linear.Coeffs.AddRange(dict.Values);
-            linear.Offset = constant;
-        }
-        else
+        if (negate)
         {
             foreach (var coeff in dict.Values)
             {
                 linear.Coeffs.Add(-coeff);
             }
             linear.Offset = -constant;
+        }
+        else
+        {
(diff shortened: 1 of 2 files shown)


ortools/sat/csharp/CpModel.cs
```diff
@@ -1191,25 +1191,36 @@ internal LinearExpr GetLinearExpr<X>(X x)
 
     internal LinearExpressionProto GetLinearExpressionProto(LinearExpr expr, bool negate = false)
     {
-        Dictionary<IntVar, long> dict = new Dictionary<IntVar, long>();
-        long constant = LinearExpr.GetVarValueMap(expr, 1L, dict, terms_);
+        var dict = var_value_map_;
```
lperron, line 1194: why introduce a local variables ?

ortools/sat/csharp/CpModel.cs
```diff
@@ -1191,25 +1191,36 @@ internal LinearExpr GetLinearExpr<X>(X x)
...
+        long constant = LinearExpr.GetVarValueMap(expr, dict, terms_);
         long mult = negate ? -1 : 1;
 
         LinearExpressionProto linear = new LinearExpressionProto();
         var dictCount = dict.Count;
         linear.Vars.Capacity = dictCount;
+        linear.Vars.AddRange(dict.Keys);
         linear.Coeffs.Capacity = dictCount;
-        foreach (KeyValuePair<IntVar, long> term in dict)
+        if (!negate)
```
lperron, line 1204: I prefer positive tests

if (negate) 

else

ortools/sat/csharp/CpModel.cs
bollhals: locals generate slightly better code due to the fact that that the JIT is able to prove e.g. that dict is not null.

ortools/sat/csharp/CpModel.cs
bollhals: fine by me, will swap

lperron:
Thanks

PR #3223, opened 2022-04-05
Update solve.cc's solver flag docs to include pdlp

mlubin:
CC @dapplegate


Mizux:
just backported to Google, Thx !

PR #3331, opened 2022-06-04
Use uintptr_t

andreas-schwab:


lperron:
Actually, it breaks our build.

I must refuse it.

andreas-schwab:
In which way??

lperron:
Ambiguous cast on Mac.

Le lun. 6 juin 2022, 13:12, Andreas Schwab ***@***.***> a
écrit :

andreas-schwab:
In which way??

PR #3370, opened 2022-06-30
Bugfix: add is_positive variable

SergeyKhapuginSmartlane:
On some systems, the vector values would be empty when resolving the second argument of the RegisterUnaryCallback call. This meant that the callback was always treated as positive. This caused issues if registering a transit vector with negative values. Creating an is_positive variable and resolving before the call fixes the issue.

P.S. Sorry for the triple PR, the CLA stuff is a bit tricky.


lperron:
Thanks for the PR.
I will redo it myself to include the correct formatting.

PR #3383, opened 2022-07-15
Fix RUNPATHs for libraries and executables

StefanBruens:
The RUNPATH should be limited to the entries actually required, instead of trying to cover all possible scenarios by adding an entry for each.

Instead of modifying the default RUNPATH by repeatedly setting CMAKE_INSTALL_RPATH, set the correct minimal INSTALL_RPATH property for each executable and library.

Only tested on Linux. Windows should not be affected, MacOS may be wrong.

### DIFF
--- a/cmake/cpp.cmake
+++ b/cmake/cpp.cmake
@@ -440,20 +440,23 @@ function(add_cxx_sample FILE_NAME)
   get_filename_component(COMPONENT_DIR ${SAMPLE_DIR} DIRECTORY)
   get_filename_component(COMPONENT_NAME ${COMPONENT_DIR} NAME)
 
-  if(APPLE)
-    set(CMAKE_INSTALL_RPATH
-      "@loader_path/../${CMAKE_INSTALL_LIBDIR};@loader_path")
-  elseif(UNIX)
-    set(CMAKE_INSTALL_RPATH
-      "$ORIGIN/../${CMAKE_INSTALL_LIBDIR}:$ORIGIN/../lib64:$ORIGIN/../lib:$ORIGIN")
-  endif()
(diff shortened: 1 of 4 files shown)


StefanBruens:
Several unrelated failures due to the recently added `python_python_line_balancing_sat` test, which is broken.

Mizux:
Overall OK for inclusion but want to perform few tests before and I prefer to postponed it to v9.5 (~August/September ?) unless you have strong argument(s) (and confidence) to include it in the incoming v9.4.

1. Will test on MacOS but since you now move the code **after** the `add_executable()` I fear it won't be take into account OR like on unix we should try instead to use a `set_target_property`
note: Will test it ASAP !

2. IIRC, We are using CMAKE_BUILD_WITH_INSTALL_RPATH (since libortools can be packaged in a "CMake orchestrated python wheel package") https://github.com/google/or-tools/blob/49b6301e1e1e231d654d79b6032e79809868a70e/CMakeLists.txt#L89-L92
And we are also using
https://github.com/google/or-tools/blob/49b6301e1e1e231d654d79b6032e79809868a70e/CMakeLists.txt#L46-L72
to try to have the same relative layout and directory naming between the buildir and the installdir i.e. `bin/` and `lib[64]/`
-> need to carefully check it is working as expected aka `CMAKE_INSTALL_LIBDIR` is the same than `CMAKE_INSTALL_FULL_LIBDIR` (relative to INSTALL_BINDIR).

StefanBruens:
Now fails to merge due to conflicting changes done in the meantime ... :-(

StefanBruens:
@Mizux Ping!

PR #3410, opened 2022-08-01
Add commentary on RCPSP example + dataset

A-Wpro:
Finding exactly what this example solve and with which data was not really easy, those comments will probably help peoples to use this example as a base for their RCPSP-like problem/project.
Commentaries are a link to help understand what is RCPSP problem and a link to dataset used to run this example.


lperron:
Thanks.

I have patched it myself because (a) we use the main branch now, and (b) I prefer a different introduction to rcpsp than the csplib that is really sketchy,.

A-Wpro:
> Thanks.
> 
> I have patched it myself because (a) we use the main branch now, and (b) I prefer a different introduction to rcpsp than the csplib that is really sketchy,.

Thank you for your answer.

PR #3415, opened 2022-08-03
Move attributes to beginning in function declarations to allow it to build on GCC 12

cactorium:
Hey there! GCC 12 apparently cares more about attribute order than GCC 11, and these two declarations caused the build to error out. It looks like just these two function declarations were misordered, at least with default build settings


Mizux:
Fixed on main few months ago
ref: https://github.com/google/or-tools/commit/672303cc2f2396c4bd5fbed8430208308815bd54

FYI `master` is deprecated and have been renamed into `main`

Hope you didn't pass too much time on it

Mizux (approved):

PR #3420, opened 2022-08-07
direct CLI example for PDLP

mlubin:
Provides an example of calling PDLP's API directly from C++.
solve.cc is a bad experience for PDLP users because of the solution validator (e.g., [this discussion](https://groups.google.com/g/or-tools-discuss/c/Cx5WRY37MZA/m/FTLXcSk6BQAJ)).
As a side effect, the SolveLog is now accessible to CLI users.

Example usage:
```
bazel run -c opt --cxxopt=-std=c++17 examples/cpp:pdlp_solve -- \
--input=/path/to/file.mps --solve_log_file=/tmp/solvelog.json --sol_file=/tmp/solve.sol
```

CC @dapplegate

### DIFF
--- a/examples/cpp/BUILD.bazel
+++ b/examples/cpp/BUILD.bazel
@@ -393,6 +393,22 @@ cc_binary(
 )
 
 cc_binary(
+    name = "pdlp_solve",
+    srcs = ["pdlp_solve.cc"],
+    deps = [
+        "//ortools/base",
+        "//ortools/pdlp:primal_dual_hybrid_gradient",
+        "//ortools/pdlp:solvers_cc_proto",
+        "//ortools/pdlp:solve_log_cc_proto",
+        "//ortools/pdlp:quadratic_program_io",
+        "//ortools/linear_solver:linear_solver_cc_proto",
+        "//ortools/port:proto_utils",
+        "//ortools/util:sigint",
+        "@com_google_absl//absl/time",
+    ],
+)
+
+cc_binary(
     name = "pdptw",
     srcs = ["pdptw.cc"],
     deps = [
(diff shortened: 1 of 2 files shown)


Mizux (approved):
LGTM

stumarcus314:
Is pdlp_solve still the preferred way to invoke PDLP rather than using the default solve binary built by cmake (according to the standard installation instructions from C++ source: https://developers.google.com/optimization/install/cpp/source_linux) and provided in ~/or-tools_main/or-tools/build/bin?

Mizux:
To be honest currently pdlp_solve.cc is not tested as well as it should.
In CMake: we currently lack of data input support so it is disabled (shame on us)
https://github.com/google/or-tools/blob/053742add375895f4f61a005cab094e24d7baa82/examples/cpp/CMakeLists.txt#L55
In Bazel: We don't test the `binary_cc()` but at least we compile it...

Concerning `linear_solver/solve.cc` at least it is build in both supported build system but again not tested as much it should be...
https://github.com/google/or-tools/blob/053742add375895f4f61a005cab094e24d7baa82/ortools/linear_solver/CMakeLists.txt#L60-L66
https://github.com/google/or-tools/blob/053742add375895f4f61a005cab094e24d7baa82/ortools/linear_solver/BUILD.bazel#L384-L386

Side note: we should also rename them so in the cmake install export it will be more clear which is which, we already start internal discussion on a loosy naming convention: `<component>_solve`.

https://github.com/google/or-tools/blob/053742add375895f4f61a005cab094e24d7baa82/ortools/sat/CMakeLists.txt#L52-L58

mlubin:
> solve.cc is a bad experience for PDLP users because of the solution validator

I believe this statement is still true. 

> Is pdlp_solve still the preferred way to invoke PDLP

It's an example of how to invoke PDLP through its direct API, though I wouldn't say it's "the preferred way". Depending on what you want to accomplish, you may want to modify and build on this example.

PR #3512, opened 2022-10-23
replace erase-remove idiom with std::erase_if

Regista6:
Reference: [std::erase, std::erase_if (std::vector)](https://en.cppreference.com/w/cpp/container/vector/erase2)


lperron:
I cannot merge it.
It requires C++20, which would be problematic.

Regista6:
Understood. Thank you for your consideration.

Issue #3503, opened 2022-10-18
Missing support for .NET Framework in latest Nuget packages


anders9ustafsson:
**What language and solver does this apply to?**
C#

**Describe the problem you are trying to solve.**
Use `Google.OrTools` Nuget package, versions 9.3 and later, in a .NET Framework 4.8 project.
 
**Describe the solution you'd like**
Being able to use the latest Nuget pacakge in .NET Framework.

**Describe alternatives you've considered**

**Additional context**
With version 9.2 and earlier, .NET Framework was supported. Why was .NET Framework support dropped?

lperron:
because .NET6 is out and is supposed to merge the two branches (dotnet code and dotnet framework).

anders9ustafsson:
> because .NET6 is out and is supposed to merge the two branches (dotnet code and dotnet framework).

Please elaborate. As far as I know, you cannot reference a .NET 6 DLL from a .NET Framework 4.x assembly.

CervEdin:
> Please elaborate. As far as I know, you cannot reference a .NET 6 DLL from a .NET Framework 4.x assembly.

@anders9ustafsson Well, you *can* but it's likely to crash.

> because .NET6 is out and is supposed to merge the two branches (dotnet code and dotnet framework).

@lperron dotnet 6 does not merge dotnet core (typo?) and dotnet framework. Tbh it's really just a rebrand of dotnet core. Dotnet standard is used as a compatibility framework to bridge the two.

anders9ustafsson:
@CervEdin Yes, it will definitely crash. First of all, you cannot include the Nuget package due to frameworks not matching. Second, if you try to reference the 9.4 Google.OrTools DLL directly, the project will indeed build, but the app will crash due to missing/mismatching DLLs.

I would like to use the PDLP linear solver, but as far as I have been able to find out it is only available in version 9.3 and onwards. The latest OR Tools Nuget package that supports .NET Framework is 9.2.

I have tested to build the stable branch of the `Google.OrTools` project from source with target framework set to *net48*. It does build, so as far as I understand it _should be_ possible to re-introduce .NET Framework support in the Nuget package.

Mizux:
My 2 cents.

## Context
0. .Net Framework `[4.x-4.6.1]` are deprecated
1. .Net Framework 4.6.2 and 4.8 still seems to be used **BUT** are not backed by Microsoft on Linux and MacOS...
   note: Mono is not well supported by dotnet cli and build ecosytem AFAIK
[...]

PR #3543, opened 2022-11-10
Replace TryDequeue with Count == 0 check

anders9ustafsson:
This PR addresses a limitation in .NET Standard 2.0, where the method `Queue.TryDequeue` is not implemented. I have refactored the associated code to instead check if the queue is empty (`Queue.Count == 0`), which should mimic the behavior of the `TryDequeue` method.

This fix is only relevant in case it is decided to target .NET Standard 2.0, see discussion in #3503 .

### DIFF
--- a/ortools/sat/csharp/CpSolver.cs
+++ b/ortools/sat/csharp/CpSolver.cs
@@ -211,10 +211,11 @@ public class CpSolver
                 throw new ArgumentException("Cannot evaluate '" + expr + "' in an integer expression");
             }
 
-            if (!terms_.TryDequeue(out var term))
+            if (terms_.Count == 0)
             {
                 break;
             }
+            var term = terms_.Dequeue();
             expr = term.expr;
             coefficient = term.coefficient;
         } while (true);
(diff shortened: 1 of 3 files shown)


CADBIMDeveloper:
Hi @anders9ustafsson , maybe it's better to check term.Count property instead of try-catch. Seems, it exists in .Net standard?

anders9ustafsson:
@CADBIMDeveloper Personally, I would say that the changes I propose in the PR are straightforward modifications of the original code, and thus preferable.

Howeer, if you are concerned with the usage of `try`-`catch`, don't hesitate to provide an alternative solution to the issue. Perhaps in a separate PR?

CADBIMDeveloper:
@anders9ustafsson I was just interested in motivation of using try-catch here. I thought I missed something about queue implementation in .Net and it was not safe to check Count property here for some reason I was not aware of.

anders9ustafsson:
@CADBIMDeveloper I will give some thought to your suggestion, you are probably right that it is a less intrusive way to use `Queue.Count` instead.

anders9ustafsson:
@CADBIMDeveloper In the end, I found your suggestion more appealing, so I have updated the pull request according to your recommendation. Many thanks!

lperron:
Closing the PR. We have a working solution

PR #3608, opened 2022-12-27
Cross Platform compatible makefiles

arnabanimesh:
Fix 'make test` command for Windows


lperron:
I applied this patch manually as we develop against the main branch and not the stable one.

Thanks

PR #3668, opened 2023-02-07
Python CP-SAT type hints

stradivari96:
Work in progress, opened as draft pull request.

Tasks:
- [x] Custom types: Integral (same logic as `cmh.is_integral`), Literal (`IntVar`+`_NotBooleanVariable`+`Integral`), Arc, etc
- [x] Run mypy on `ortools/sat/samples`

### DIFF
--- a/ortools/sat/python/cp_model.py
+++ b/ortools/sat/python/cp_model.py
@@ -46,10 +46,13 @@ rather than for solving specific optimization problems.
 """
 
 import collections
+import numbers
 import threading
 import time
-from typing import Optional
+from typing import Optional, Iterable, Union, Tuple, Sequence, overload
 import warnings
+
+import numpy as np
 
 from ortools.sat import cp_model_pb2
 from ortools.sat import sat_parameters_pb2
@@ -93,6 +96,13 @@ AUTOMATIC_SEARCH = sat_parameters_pb2.SatParameters.AUTOMATIC_SEARCH
 FIXED_SEARCH = sat_parameters_pb2.SatParameters.FIXED_SEARCH
 PORTFOLIO_SEARCH = sat_parameters_pb2.SatParameters.PORTFOLIO_SEARCH
 LP_SEARCH = sat_parameters_pb2.SatParameters.LP_SEARCH
+
+# Type aliases
+IntegralT = Union[numbers.Integral, np.integer, int]
+LiteralT = Union["IntVar", "_NotBooleanVariable", IntegralT, bool]
+VariableT = Union["IntVar", "_ProductCst", IntegralT]
+LinearExprT = Union["LinearExpr", "IntVar", "_ProductCst", IntegralT]
+ArcT = Tuple[IntegralT, IntegralT, LiteralT]
 
 
 def DisplayBounds(bounds):
@@ -144,7 +154,7 @@ def ShortExprName(model, e):
     return str(e)
 
 
-class LinearExpr(object):
+class LinearExpr:
     """Holds an integer linear expression.
 
     A linear expression is built from integer constants and variables.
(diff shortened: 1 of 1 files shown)


ortools/sat/python/cp_model.py
```diff
@@ -956,31 +959,32 @@ def ObjectIsAFalseLiteral(literal):
...
 
-    def SetName(self, name):
+    def SetName(self, name: str):
         """Sets the name of the model."""
         self.__model.name = name
 
     # Integer variable.
 
-    def NewIntVar(self, lb, ub, name):
+    def NewIntVar(self, lb, ub, name: str) -> IntVar:
```
oulianov, line 987: `def NewIntVar(self, lb: Union[float, IntVar], ub: Union[float, IntVar], name: str) -> IntVar:`

ortools/sat/python/cp_model.py
```diff
@@ -1647,7 +1651,7 @@ def AddMultiplicationEquality(self, target, *expressions):
 
     # Scheduling support
 
-    def NewIntervalVar(self, start, size, end, name):
+    def NewIntervalVar(self, start, size, end, name: str) -> IntervalVar:
```
oulianov, line 1654: `def NewIntervalVar(self, start: Union[int, IntVar], size: Union[int, IntVar], end: Union[int, IntVar], name: str) -> IntervalVar:`

ortools/sat/python/cp_model.py
```diff
@@ -1634,7 +1638,7 @@ def AddModuloEquality(self, target, var, mod):
         model_ct.int_mod.target.CopyFrom(self.ParseLinearExpression(target))
         return ct
 
-    def AddMultiplicationEquality(self, target, *expressions):
+    def AddMultiplicationEquality(self, target, *expressions) -> Constraint:
         """Adds `target == expressions[0] * .. * expressions[n]`."""
```
oulianov, line 1642: Should we correct this to reflect the fact that MultiplicationEquality will fail if `len(expressions) > 2` ?

ortools/sat/python/cp_model.py
```diff
@@ -1824,7 +1828,7 @@ def AddNoOverlap2D(self, x_intervals, y_intervals):
             [self.GetIntervalIndex(x) for x in y_intervals])
         return ct
 
-    def AddCumulative(self, intervals, demands, capacity):
+    def AddCumulative(self, intervals: Iterable[IntervalVar], demands, capacity) -> Constraint:
```
oulianov, line 1831: `def AddCumulative(self, intervals: Iterable[IntervalVar], demands: Iterable[Union[int, IntVar]], capacity: Union[int, IntVar]) -> Constraint:`

oulianov:
Thank you for opening this review ! That's a great start.

I added some comments right now with additional typing from the top of my mind. 

What would be the best way to help you cover all the definitions in the file ?

ortools/sat/python/cp_model.py
stradivari96: I'll probably define it as:
```
IntegralT = Union[numbers.Integral, np.integer]
```
See:
https://github.com/google/or-tools/blob/8179abf8346b02c9aa39d356177fdbd6315f2c34/ortools/sat/python/cp_model_helper.py#L24-L26

stradivari96:
Thanks for the help offer and suggestions! @oulianov 

But I believe I just need some time to figure out the best type definitions for some of the more complex interfaces, eg:
- OnlyEnforceIf
- arcs (iterable of sequence of len 3)
- etc

And test them thoroughly.

oulianov:
For sure, totally understand. Wish you the best, feel free to reach out

ortools/sat/python/cp_model.py
lperron: At some point, we will support it. So I would say no.

lperron:
Hi, 

Any change of pushing this ? 
Just run the black formatter on the code before committing.

stradivari96:
My bad, never saw the conflicts, should be ready to review/merge now

PR #3719, opened 2023-03-26
Fix reference to solve_wrapper in Solve method

djsilcock:
Calling `solver.StopSearch()` (as opposed to from within the solution callback) does not work:

    solver=CPSolver()
    def stopper():
        solver.StopSearch()
    threading.Timer(30,stopper) #in real life would be some external event
    solver.Solve()

because self.__solve_wrapper is never set in Solve method


lperron:
Thanks

PR #3728, opened 2023-03-30
ensure absolute native library path in Java Loader

afloren:
Ensures that the native library path is absolute before loading to avoid `java.lang.UnsatisfiedLinkError: Expecting an absolute path of the library`.

---------

In my particular case the issue arises when running as part of an application that has `java.io.tmpdir` set to a relative path which results in the constructed tempDir path being relative. We could alternatively force the tempDir path to be absolute immediately on creation, but forcing the path to be absolute immediately before the call that cares seems more future proof. I'm currently using a custom loader to avoid this issue but I figured I'd contribute in case anyone else hits this edge case.

### DIFF
--- a/ortools/java/com/google/ortools/Loader.java
+++ b/ortools/java/com/google/ortools/Loader.java
@@ -120,6 +120,7 @@ public class Loader {
         // Load the native library
         System.load(tempPath.resolve(RESOURCE_PATH)
                         .resolve(System.mapLibraryName("jniortools"))
+                        .toAbsolutePath()
                         .toString());
         loaded = true;
       } catch (IOException e) {


afloren:
happy to add a test case for the offending tmpdir setting as well though I thought it might be overkill

Mizux (approved):

afloren:
Hey @Mizux thanks for reviewing! I don't think the checks issues are related to the PR content, is there anything you'd like me to do before merging?

Mizux:
> Hey @Mizux thanks for reviewing! I don't think the checks issues are related to the PR content, is there anything you'd like me to do before merging?

Yup, as usual our main (aka dev) branch is borken on windows...

To be honest, I just wanted to play with this patch to understand all the implications (e.g. try to use it in a maven Java app to see what's happen, test it in https://github.com/mizux/java-native etc...).
I guess we have to follow plan B -> I'll merge it and revert it if lot of user complains about some path error/loading issue.

Sorry for the delay and thx for the PR.

Issue #3743, opened 2023-04-06
Gurobi environment errors are not shown


ianr15:
**What language and solver does this apply to?**
All languages, linear solver / Gurobi

**Describe the problem you are trying to solve.**
Gurobi errors arising when loading the Gurobi environment are not logged in OR-Tools, which makes troubleshooting hard.

**Describe the solution you'd like**
Logging of Gurobi loading errors

**Describe alternatives you've considered**
NA

**Additional context**
Based on github.com/google/or-tools/discussions/3739

ianr15:
I will make a PR to fix this (I already have a local installation of OR-Tools doing it)

ianr15:
https://github.com/google/or-tools/pull/3744

lperron:
Here is the output I gen now:

Found the Gurobi library in '/Library/gurobi1000/macos_universal2/lib/libgurobi100.dylib.
FAILED_PRECONDITION: Found the Gurobi shared library, but could not create Gurobi environment: is Gurobi licensed on this machine?No Gurobi license found

PR #3744, opened 2023-04-06
Issue #3743: add Gurobi error log

ianr15:
Log Gurobi error if Gurobi environment cannot be loaded


lperron:
Wrong place. It should be in the function at the top of the file.
And PR should be against main (and not stable).

I will look.

lperron:
I have pushed my code to do this.

Mizux:
https://github.com/google/or-tools/commit/dfdf042f6be3f9612c44f58be1035954024847e3#diff-c66dfba5003e496974e5afdd9cd57908ea5416e389f5c8396afbd3576c583731R35

PR #3831, opened 2023-06-22
feat(python): unignore Solver::TopProgressPercent and SearchMonitor::…

bertuccellimatteo:
…Progress


lperron:
Note that the actual numbers if very approximate and not very useful IMO.

PR #3837, opened 2023-06-28
Evaluate LinearExp for the found solution

koen-lee:
This PR adds a `SolutionValue()` method to the `LinearExp` class, similar to the `SolutionValue()` method of `Variable`.
This allows easy reconstruction of all partial constraints when a solution is found.


lperron:
Nice

PR #3845, opened 2023-07-04
Fix CMake build in `test` subdirectories

lnikkila:
Due to the regexes, ortools_pdlp fails to build if any of its parent directories starts with `test` or `gtest`:

    CMake Error at test/ortools-src/ortools/pdlp/CMakeLists.txt:27 (add_library):
      No SOURCES given to target: ortools_pdlp


Mizux (approved):
Good Catch and thx for the fix !

Issue #4010, opened 2023-12-02
file::GetContents and file::SetContents leak memory


pjh:
**What version of OR-Tools and what language are you using?**
Version: stable, at head as of today
Language: C++

**Which solver are you using (e.g. CP-SAT, Routing Solver, GLOP, BOP, Gurobi)**
n/a

**What operating system (Linux, Windows, ...) and version?**
Linux x86_64

**What did you do?**
I'm using the `//ortools/base:file` library and with valgrind I detected a memory leak when using `file::GetContents` or `file::SetContents`. I've created https://github.com/google/or-tools/pull/4009 to demonstrate the leak with a minimal example.

**What did you expect to see**
No memory leaks

**What did you see instead?**
"definitely lost: 48 bytes in 2 blocks"

lperron:
Thanks. Merged.

PR #4009, opened 2023-12-02
Demonstrate file::GetContents / SetContents memory leak

pjh:
To reproduce the memory leak cherry-pick this PR, ensure you have `valgrind` installed and then:

```
bazel build //ortools/base:hello_file
valgrind --leak-check=full ./bazel-bin/ortools/base/hello_file &> leak.out
bazel build --cxxopt=-DFIX_MEMORY_LEAK //ortools/base:hello_file
valgrind --leak-check=full ./bazel-bin/ortools/base/hello_file &> no-leak.out
```

Here's the output I see: [leak.out](https://gist.github.com/pjh/5026217ec5faa84c45bd277d57af7bea), [no-leak.out](https://gist.github.com/pjh/42267ea20493b1e8afe0f1342458cfe5)

valgrind reports "definitely lost: 48 bytes in 2 blocks" for or-tools at head, but "definitely lost: 0 bytes in 0 blocks" when the File is deleted after closing it.

PR #4011, opened 2023-12-02
Fix memory leaks in file::GetContents and file::SetContents

pjh:
Calls `delete file` after closing opened files in `file::GetContents` and `file::SetContents`. Documents the need to do so for callers of `file::Open` and its variants.

Fixes https://github.com/google/or-tools/issues/4010

I've verified this fix using the minimal `hello_file` example in https://github.com/google/or-tools/pull/4009 but reviewers please let me know how else this should be tested.


pjh:
cc @rachelcao-g

PR #4020, opened 2023-12-11
Publish handler for user-defined MPSolverInterface

flomnes:
The goal of this work is to allow the user to use a custom `MPSolverInterface`, without having their own (forked / re-compiled) version of OR-Tools. They could avoid using the current BuildSolverInterface and it's large `switch..case` statement.

For example we have a solver (sirius) that don't want to include into the main repository here. We would just need a small library exposing a `BuildSiriusInterface` function on our side.

### DIFF
--- a/ortools/linear_solver/linear_solver.h
+++ b/ortools/linear_solver/linear_solver.h
@@ -188,6 +188,8 @@ bool SolverTypeIsMip(MPModelRequest::SolverType solver_type);
  */
 class MPSolver {
  public:
+  // Advanced usage : custom interface builder
+  using MPSolverInterfaceBuilder = MPSolverInterface*(*)(MPSolver* const);
   /**
    * The type of problems (LP or MIP) that will be solved and the underlying
    *  solver (GLOP, GLPK, CLP, CBC or SCIP) that will solve them. This must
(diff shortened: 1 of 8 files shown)


ortools/linear_solver/linear_solver.h
```diff
@@ -188,6 +188,8 @@ bool SolverTypeIsMip(MPModelRequest::SolverType solver_type);
  */
 class MPSolver {
  public:
+  // Advanced usage : custom interface builder
+  using MPSolverInterfaceBuilder = MPSolverInterface*(*)(MPSolver* const);
```
flomnes, line 192: Do we need an interface class here instead of a C-style function pointer ?

lperron:
Knowing this: https://github.com/google/or-tools/discussions/3992

I am not sure we should add more feature for MPSolver. I would prefer we concentrate on ModelBuilder and MathOpt by re-using all the blocks of the MPSolver support for XPressMP.

In particular, all basis optimization should be done while discussing with the math-opt team.

flomnes:
To be tested with Sirius

flomnes:
Mixed results, issues with privacy (`friend` classes & functions)

PR #4041, opened 2024-01-04
Update shift_scheduling_sat.py; adjusting comments

krzysztofkrzysztof:
Two omments on python scheduling example are misleading


Mizux:
line 383 Minimize() i.e. negative value reduce cost so incentive the solver to take it i.e. want

PR #4048, opened 2024-01-10
Include LICENSE file in Python distribution

chrisjbremner:
I have tooling in place to pull the license text from packages using [`pip-licenses`](https://github.com/raimon49/pip-licenses). I noticed, however, that `ortools` does not include the license file in its Python distribution, and as a result, it is difficult to pull in the license file from the project, especially since there is no link back to the Github repo in the Python package (`https://developers.google.com/optimization/` is listed as the package URL).

I am proposing including the `LICENSE` file in the Python distribution by copying the License in as part of the `python.cmake` file. By doing so, it will get picked up by `setuptools` since it defaults to including files that match the following patterns: `'LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*'` as described in code [here](https://github.com/pypa/setuptools/blob/8533749d7f009e2e517791ed91ff36f7d36aa3fb/setuptools/dist.py#L384-L401) or in the docs [here](https://setuptools.pypa.io/en/latest/references/keywords.html#keywords) (under `license_files`).

In addition, I removed the `setup_data.py` since it seemed like an outdated version of `setup.py.in` which appears to be what has been used recently. Please correct me if I shouldn't have removed this.

I will note that I have yet to be able to test this since I don't have the full toolchain set up.


Mizux:
Also in pyproject.toml (since `setup.py` seems deprecated AFAIK)
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

chrisjbremner:
Just checking -- is that a request to convert to using pyproject.toml for packaging instead of setup.py in this PR, or are you simply pointing out that doing this in pyproject.toml is a possibility in case that's used in this project in the future?

Mizux:
> Just checking -- is that a request to convert to using pyproject.toml for packaging instead of setup.py in this PR, or are you simply pointing out that doing this in pyproject.toml is a possibility in case that's used in this project in the future?

AFAIK setup.py is slowly being deprecated so it is good to know that this LICENSE stuff is also compliant with the toml approach so the day we [have to] migrate we already have a clean shape.

PR #4081, opened 2024-02-12
Fix typo in log message

ubersan:
Tiny typo in the log message fixed


lperron:
I pushed it myself as it needs to target main and not stable

Thanks

ubersan:
ah yea my bad, thx

Issue #4120, opened 2024-03-09
python type annotations too incomplete to be useful?


dimbleby:
python, version 9.9.3963

this release includes some `py.typed` files.  I like to add annotations to my python code and check them strictly, so I am totally in favour of the idea here.

however, the experience of the annotations as they are today is awkward - they generate lots of false positives on simple code.  

Here are the cases that I have found myself having to `# type: ignore`:

```python
from ortools.sat.python import cp_model
model = cp_model.CpModel()
```
result: `error: Call to untyped function "CpModel" in typed context  [no-untyped-call]`

```python
from ortools.sat.python import cp_model
solver = cp_model.CpSolver()
```
result: `error: Call to untyped function "CpSolver" in typed context  [no-untyped-call]`
```python
      status = solver.Solve(model)
      if status in (cp_model.FEASIBLE, cp_model.OPTIMAL):
          pass
```
result: `error: Non-overlapping container check (element type: "CpSolverStatus", container item type: "ValueType")  [comparison-overlap]`
```
  class SolutionPrinter(cp_model.CpSolverSolutionCallback):
      def __init__(self):
         super().__init__()
```
result: `Call to untyped function "__init__" in typed context  [no-untyped-call]`

please consider either improving the annotations to make them more useful, else not publishing `py.typed` until they are more ready for consumption.

dimbleby:
I think three out of those four can be resolved by adding ` -> None` to `__init__()` methods

dimbleby:
Another one:
```python
      model = cp_model.CpModel()  # type: ignore[no-untyped-call]
      foo = model.NewIntVar(1, 26, "foo")
      bar = model.NewIntVar(1, 26, "bar")
      baz = model.NewIntVar(1, 26, "baz")
      quux = model.NewIntVar(1, 26, "quux")
      thud = model.NewIntVar(1, 26, "thud")
     model.Add(quux == foo + bar + baz + thud)
```
result: 
```
foo.py:14: error: No overload variant of "__radd__" of "LinearExpr" matches argument type "SupportsComplex"  [operator]
foo.py:14: note: Possible overload variants:
foo.py:14: note:     def __radd__(self, LinearExpr | IntVar | Integral | int, /) -> LinearExpr | IntVar | Integral | int
foo.py:14: note:     def __radd__(self, LinearExpr | IntVar | Integral | int | Number | float, /) -> LinearExpr | IntVar | Integral | int | Number | float
[...]

PR #4121, opened 2024-03-09
some fixes to the python typing

dimbleby:
fixes the specific complaints in #4120

if there are others then I leave it to you / other contributors to find and fix them

wibni CI pipelines executed typechecking so as to find such things automatically, but figuring out how that all would fit together is more than I am willing to offer


dimbleby:
oh a CLA.  Sorry, I won't be getting into that.

Feel free to take these fixes, or not.

lperron:
Thanks

PR #4150, opened 2024-03-22
Use fixtures for XPRESS tests, instead of C-style macros

flomnes:
Instead of macros, use fixtures for XPRESS tests.


pet-mit (approved):

lperron:
can you re-create against the main branch ?

flomnes:
> can you re-create against the main branch ?

Done

PR #4184, opened 2024-04-15
[XPRESS] Extract variable & constraint names only before writing MPS + fix a test

pet-mit:
- Fix a unit test in XPRESS interface (it was failing because of difficult-to-handle trailing spaces)
- XPRESS interface: extract variable & constraint names, only before writing MPS, in order to have an MPS with clear names while avoiding extracting the names too often (extraction in XPRESS is costly)


pet-mit:
CLA issue fixed in new commit : https://github.com/google/or-tools/pull/4185

PR #4185, opened 2024-04-15
[XPRESS] Extract variable & constraint names only before writing MPS + fix a test

pet-mit:
- Fix a unit test in XPRESS interface (it was failing because of difficult-to-handle trailing spaces)
- XPRESS interface: extract variable & constraint names, only before writing MPS, in order to have an MPS with clear names while avoiding extracting the names too often (extraction in XPRESS is costly)

### DIFF
--- a/ortools/linear_solver/xpress_interface.cc
+++ b/ortools/linear_solver/xpress_interface.cc
@@ -2056,7 +2056,6 @@ template <class T>
 // T = MPVariable | MPConstraint
 // or any class that has a public method name() const
 void ExtractNames(XPRSprob mLp, const std::vector<T*>& objects) {
-  MPObjective o;
   const bool have_names =
       std::any_of(objects.begin(), objects.end(),
                   [](const T* x) { return !x->name().empty(); });
--- a/ortools/linear_solver/xpress_interface.cc
+++ b/ortools/linear_solver/xpress_interface.cc
@@ -277,12 +277,13 @@ class MPCallbackWrapper {
     for (const std::exception_ptr& ex : caught_exceptions_) {
       try {
         std::rethrow_exception(ex);
-      } catch (std::exception &ex) {
+      } catch (std::exception& ex) {
         // We don't want the interface to throw exceptions, plus it causes
         // SWIG issues in Java & Python. Instead, we'll only log them.
         // (The use cases where the user has to raise an exception inside their
         // call-back does not seem to be frequent, anyway.)
-        LOG(ERROR) << "Caught exception during user-defined call-back: " << ex.what();
+        LOG(ERROR) << "Caught exception during user-defined call-back: "
+                   << ex.what();
       }
     }
     caught_exceptions_.clear();
@@ -1243,8 +1244,7 @@ int64_t XpressInterface::nodes() const {
 }
 
 // Transform a XPRESS basis status to an MPSolver basis status.
-MPSolver::BasisStatus XpressToMPSolverBasisStatus(
-    int xpress_basis_status) {
+MPSolver::BasisStatus XpressToMPSolverBasisStatus(int xpress_basis_status) {
   switch (xpress_basis_status) {
     case XPRS_AT_LOWER:
       return MPSolver::AT_LOWER_BOUND;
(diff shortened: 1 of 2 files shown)


ortools/linear_solver/xpress_interface.cc
```diff
@@ -277,13 +277,12 @@ class MPCallbackWrapper {
     for (const std::exception_ptr& ex : caught_exceptions_) {
       try {
         std::rethrow_exception(ex);
-      } catch (std::exception& ex) {
+      } catch (std::exception &ex) {
```
Mizux, line 280: Could you try to run the following command on your PR to fix formatting ?
`clang-format --style=file -i ortools/linear_solver/xpress_interface.cc`

ref: https://packages.ubuntu.com/jammy/clang-format

ortools/linear_solver/xpress_interface.cc
```diff
@@ -2045,11 +2036,60 @@ MPSolver::ResultStatus XpressInterface::Solve(MPSolverParameters const& param) {
...
+}  // namespace
+
 void XpressInterface::Write(const std::string& filename) {
   if (sync_status_ == MUST_RELOAD) {
     Reset();
   }
   ExtractModel();
+
+  ExtractNames(mLp, solver_->constraints_);
+  ExtractNames(mLp, solver_->variables_);
```
Mizux, line 2091: nit: I would extract variables before extracting constraints

ortools/linear_solver/xpress_interface.cc
pet-mit: sure, must I do it for the test class too (it would make the new test less readable)?

ortools/linear_solver/xpress_interface.cc
```diff
@@ -2045,11 +2036,60 @@ MPSolver::ResultStatus XpressInterface::Solve(MPSolverParameters const& param) {
...
+template <>
+struct getNameFlag<MPVariable> {
+  enum { value = XPRS_NAMES_COLUMN };
+};
+
+template <>
+struct getNameFlag<MPConstraint> {
+  enum { value = XPRS_NAMES_ROW };
+};
+
```
Mizux, line 2052: Should will also do the same  for `MPObjective` ?
aka XPRS_NAMES_OBJECTIVE

pet-mit: We could, but it seems unnecessary: we wouldn't use it, since MPObjective does not have a name

ortools/linear_solver/xpress_interface_test.cc
```diff
@@ -730,30 +730,24 @@ TEST_F(XpressFixtureMIP, Write) {
...
-RANGES
-    RNG00001  R2        2
-BOUNDS
- UI BND00001  C1        9
- LO BND00001  C1        -1
- UP BND00001  C2        5.147593849384714
- LO BND00001  C2        -1
-ENDATA
-)");
+  std::string expectedMps =
```
Mizux, line 733: Why not keeping the Raw string literal `R"` ?
Personally, I feel it more easier to read (thant splitting line by line with all `+ "\n"` at the end), or when you want to diff terminal output against the source file

note you can also use `// clang-format off` / `// clang-format on` to disable/enable formatting on a section of the code (e.g. data matrix manually formatted).

pet-mit: we had to change this because we saw that sometimes trailing spaces at the end of lines were automatically removed by IDEs (for example, the `" L  R1              "` line), but they're actually part of the expected MPS. If you prefer, I can reduce this change to the impacted lines.

Mizux: OK keep it like you have done !

Mizux (approved):
LGTM

Mizux:
`static int MPSolverToXpressBasisStatus(` line 486 there is also one...

pet-mit:
> `static int MPSolverToXpressBasisStatus(` line 486 there is also one...

I'm sorry I don't understand

ortools/linear_solver/xpress_interface.cc
```diff
@@ -1244,8 +1244,7 @@ int64_t XpressInterface::nodes() const {
 }
 
 // Transform a XPRESS basis status to an MPSolver basis status.
-static MPSolver::BasisStatus XpressToMPSolverBasisStatus(
-    int xpress_basis_status) {
+MPSolver::BasisStatus XpressToMPSolverBasisStatus(int xpress_basis_status) {
```
Mizux, line 1247: this method is also declared line 489 above so if removing the static, you should change the prototype there too imho.

Mizux:
> > `static int MPSolverToXpressBasisStatus(` line 486 there is also one...
> 
> I'm sorry I don't understand

my bad it was about the prototype change of `static MPSolver::BasisStatus XpressToMPSolverBasisStatus` but what you do is ok, sorry for the noise

PR #4214, opened 2024-05-02
Fix xpress range constraint extraction

djunglas:
This fixes a bug in the extraction of ranged constraints in the Xpress connector.
The problem was that the `haveRanges` flag was not reset at the beginning of the inner loop in `ExtractNewConstraints()`. This means that as soon as we have seen a chunk with a ranged constraint, all subsequent chunks were considered to have range constraints as well. Moreover, calling XPRSchgrhsrange() turns a non-ranged row into a ranged row even if the range value is 0. This causes "greater than or equal" constraints to turn into "equals" constraints.
As it turns out, calling `XPRSchgrhsrange()` is not necessary at all. Range constraints are already setup properly when callback `XPRSaddrows()`. So the easy fix is to just remove the calls to `XPRSchgrhsrange()` and everything that was setup only for this call.


pet-mit:
Wow this was a big one, nice catch! I just spent a few days debugging our software with v9.9, where it seems we met this issue. It's a shame I hadn't seen your PR. Thanks for this Daniel! :)

PR #4256, opened 2024-06-04
Update fp_utils.h for OpenBSD compatibility

ron-at-swgy:
The `fenv` struct is similar in shape to that of FreeBSD. For reference, the [official CVSWeb repository](https://cvsweb.openbsd.org/src/sys/arch/amd64/include/fenv.h?rev=1.4&content-type=text/x-cvsweb-markup) shows a definition of:

```c
/*
 * fenv_t represents the entire floating-point environment.
 */
typedef	struct {
	struct {
		unsigned int __control;		/* Control word register */
		unsigned int __status;		/* Status word register */
		unsigned int __tag;		/* Tag word register */
		unsigned int __others[4];	/* EIP, Pointer Selector, etc */
	} __x87;
	unsigned int __mxcsr;			/* Control, status register */
} fenv_t;
``` 

This is in alignment with that of the FreeBSD project.

### DIFF
--- a/ortools/util/fp_utils.h
+++ b/ortools/util/fp_utils.h
@@ -89,7 +89,7 @@ class ScopedFloatingPointEnv {
     excepts &= FE_ALL_EXCEPT;
 #if defined(__APPLE__)
     fenv_.__control &= ~excepts;
-#elif defined(__FreeBSD__)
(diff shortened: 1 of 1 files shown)


ron-at-swgy:
I've got a few other changes that can be rolled into one... stand by

Issue #4234, opened 2024-05-14
VRP with breaks minor bug?


vuongdanghuy:
**What version of OR-Tools and what language are you using?**
Version: stable branch
Language: C++

**Which solver are you using (e.g. CP-SAT, Routing Solver, GLOP, BOP, Gurobi)**
Routing Library

**What operating system (Linux, Windows, ...) and version?**
Linux Mint 21.3

**What did you do?**
Using VRP with breaks. `GlobalVehicleBreaksConstraint` will be added to the Solver.

**What did you expect to see**
I believe here we're updating `StartMin` value of the arc `path_[pos] -> path_[pos + 1]` to `EndMin` value of the break interval. This arc `StartMin` is `CumulVar(path_[pos])->Min() - arc_start_offset`. So instead of `SetMin(CapSub(interval_end_min, arc_start_offset))` it should be `SetMin(CapAdd(interval_end_min, arc_start_offset))`.

https://github.com/google/or-tools/blob/5124f4975bd6ab5a863ea55657833ab640730429/ortools/constraint_solver/routing_breaks.cc#L943-L951

stevengaygoogle:
You are correct, this should be a CapAdd() instead of a CapSub().
The current SetMin() is weaker than it could be.
Thank you for the report!

varshneydevansh:
So all we have to do is to replace the **CapSub** with the **CapAdd** ?

Or is there anything else which I should consider?

vuongdanghuy:
yeah, replace CapSub by CapAdd will do

varshneydevansh:
Okay creating a PR :)

PR #4261, opened 2024-06-05
replace CapSub with CapAdd for StartMin value

varshneydevansh:
### What this PR fixes?
Fix VRP with Breaks Constraint Calculation in Routing Solver.
- #4234 

### Explanation -

This PR solves a minor bug where we had **CapSub** instead of **CapAdd**.

The incorrect use of CapSub resulted in a weaker constraint, potentially allowing unrealistic scheduling in vehicle routing problems with breaks. By using CapAdd, we correctly adjust the start time after a break, ensuring that the solver's constraints reflect actual break intervals more accurately.


lperron:
Hi. This is fixed on main.

PR #4266, opened 2024-06-07
Diagnostic addition to cmake file

ron-at-swgy:
This commit adds an OpenBSD specific flag for use with SWIG,
`-DSMALL_LONG`, and further specifies the `-DSWIGWORDSIZE` flag
conditions to exclude OpenBSD (as is done with Apple).

## Validation of CMake platform check:

```cmake
message(STATUS "system name: ${CMAKE_SYSTEM_NAME}")

if(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
  message(STATUS "On OpenBSD")
  list(APPEND CMAKE_SWIG_FLAGS "-DSMALL_LONG")
else()
  message(STATUS "NOT on OpenBSD")
endif()

message(FATAL_ERROR "stop cmake configure to fast iterate ! (to remove after dev/debug)")
```

```shell
-- system name: OpenBSD
-- On OpenBSD
CMake Error at cmake/python.cmake:43 (message):
  stop cmake configure to fast iterate ! (to remove after dev/debug)
  Call Stack (most recent call first):
    CMakeLists.txt:429 (include)
```
## Result

This change, combined with #4259 and #4257 allows me to build the `.whl` file for
the python `or-tools` package. I then validated the `CBC_MIXED_INTEGER_PROGRAMMING`
and `GLOP` solvers with a simple linear programming problem in the Jupyter Lab
environment.

```shell
Installing collected packages: immutabledict, absl-py, ortools
Successfully installed absl-py-2.1.0 immutabledict-4.2.0 ortools-9.10.9999
```


cmake/python.cmake
```diff
@@ -31,7 +31,12 @@ if(${SWIG_VERSION} VERSION_GREATER_EQUAL 4)
   list(APPEND CMAKE_SWIG_FLAGS "-doxygen")
 endif()
 
-if(UNIX AND NOT APPLE)
+# On OpenBSD, int64_t is long long int
+if(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
+  list(APPEND CMAKE_SWIG_FLAGS "-DSMALL_LONG")
+endif()
+
+if(UNIX AND NOT APPLE AND NOT (CMAKE_SYSTEM_NAME STREQUAL "OpenBSD"))
```
ron-at-swgy, line 39: The parens around `CMAKE_SYSTEM_NAME STREQUAL "OpenBSD"` may be superfluous. I'm not familiar with operator associativity rules for CMake files.

cmake/python.cmake
```diff
@@ -31,7 +31,12 @@ if(${SWIG_VERSION} VERSION_GREATER_EQUAL 4)
   list(APPEND CMAKE_SWIG_FLAGS "-doxygen")
 endif()
 
-if(UNIX AND NOT APPLE)
+# On OpenBSD, int64_t is long long int
+if(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
+  list(APPEND CMAKE_SWIG_FLAGS "-DSMALL_LONG")
```
Mizux, line 36: not sure you need this. `SMALL_LONG` is not part of the SWIG implem nor or-tools to have `long long int` simply **NOT** define `-DSWIGWORDSIZE64` (like you did below adding the "AND NOT OpenBSD")
ref: https://github.com/swig/swig/blob/v4.2.1/Lib/stdint.i

note: The only reference to `SMALL_LONG` was in https://github.com/swig/swig/issues/1557


side note: for Java and C# with fix the typemap here
https://github.com/google/or-tools/blob/main/ortools/base/base.i

cmake/python.cmake
Mizux: like as it, better readability...

cmake/python.cmake
ron-at-swgy: I will kick off a build with the "SMALL_LONG" flag removed. Will report back

PR #4274, opened 2024-06-15
Permutation flow shop example

leonlan:
This PR adds an example of the permutation flow shop problem.

I came across [this](https://groups.google.com/g/or-tools-discuss/c/6O92_XGOrak) discussion on the Google discussion board. Based on the explanation there by @lperron I wrote this model and figured it would be helpful to share it with others :-).

I couldn't find any guidelines for submitting examples, so please let me know if this is OK!

### DIFF
--- a/examples/contrib/permutation_flow_shop.py
+++ b/examples/contrib/permutation_flow_shop.py
@@ -1,0 +1,177 @@
+#!/usr/bin/env python3
+# Copyright 2010-2024 Google LLC
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


examples/contrib/permutation_flow_shop.py
```diff
@@ -0,0 +1,177 @@
...
+def parse_args():
+    parser = argparse.ArgumentParser("Solve a permutation flow shop problem.")
+
+    msg = "Time limit in seconds. Default is no time limit."
+    parser.add_argument(
+        "--time-limit", type=float, default=float("inf"), help=msg
+    )
+
+    msg = "Whether to log the solver output."
+    parser.add_argument("--log", action="store_true", help=msg)
```
lperron, line 152: I would add a --params argument to pass additional parameters to the solver.

PR #4287, opened 2024-06-27
Update tsp_cities.py

bnjf:


Mizux:
Many thanks for this PR but our dev branch is `main`, also there is the same issue in tsp_circuit_board so I did this commit
to fix both on the correct branch..
https://github.com/google/or-tools/commit/cf94c2dc94c494c87840856b2c1a281e0534b940

Issue #4132, opened 2024-03-15
static assertion failed: ERROR_ConstraintInfo_is_not_well_compacted with MinGW


Sedfer:
Compilation failed on stable (3ed2b2d963014f285108058a7354b0afd9d81c00):
```
ortools/sat/linear_propagation.h:214:40: error: static assertion failed: ERROR_ConstraintInfo_is_not_well_compacted
  214 |   static_assert(sizeof(ConstraintInfo) == 24,
      |                 ~~~~~~~~~~~~~~~~~~~~~~~^~~~~
```

I am using Windows 10 64-bit with MinGW-w64 9.0

Looks like there is a problem with bit fields and alignment, although I am not sure why it is windows-specific. First 3 fields occupy 4 bytes each and total struct size is 32 bytes.
```c++
struct ConstraintInfo {
  unsigned int enf_status : 2;
  bool all_coeffs_are_one : 1;
  unsigned int initial_size : 29;  // Const. The size including all terms.

  EnforcementId enf_id;  // Const. The id in enforcement_propagator_.
  int start;             // Const. The start of the constraint in the buffers.
  int rev_size;          // The size of the non-fixed terms.
  IntegerValue rev_rhs;  // The current rhs, updated on fixed terms.
};

#if !defined(_MSC_VER)
  static_assert(sizeof(ConstraintInfo) == 24,
                "ERROR_ConstraintInfo_is_not_well_compacted");
#endif  // !defined(_MSC_VER)
```

I was able to fix this by simply changing second field to `unsigned int all_coeffs_are_one : 1;`.

Alternatively, it can be suppressed with `!defined(__MINGW32__) && !defined(__MINGW64__)`, just like c0e5ca1cd107de803e358528e8023e2626fa15a8 did, but I am not really sure what was the point.

lperron:
So, we do not support mingw.

Now, if you submit a PR that fixes this (change the #if !defined test to also rule out mingw, we will accept it).

PR #4293, opened 2024-07-02
Disable well compacted assert on MinGW

bonukai:
MinGW build fails, as in #4132. This pull requests implements removal of assertion for MinGW, like suggested in the linked issue.

Alternatively, `bool all_coeffs_are_one : 1` could be changed to `unsigned int all_coeffs_are_one : 1`. With this change, both MSVC and MinGW will pack the `ConstraintInfo` struct to expected 24 bytes.

### DIFF
--- a/ortools/sat/linear_propagation.h
+++ b/ortools/sat/linear_propagation.h
@@ -332,10 +332,10 @@ class LinearPropagator : public PropagatorInterface, ReversibleInterface {
     IntegerValue rev_rhs;  // The current rhs, updated on fixed terms.
   };
 
-#if !defined(_MSC_VER)
+#if !defined(_MSC_VER) && !defined(__MINGW32__) && !defined(__MINGW64__)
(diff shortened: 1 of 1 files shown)


legrosbuffle:
This assert [should also fail](https://godbolt.org/z/oro4cfGoW) with MSVC as this is the default way to pack fields on windows. Other alternatives include :

 - building with [`-mno-ms-bitfields`](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html#x86-Options), which will pack like clang, but then the binary won't be ABI compatible with others built with MSVC
 - marking with struct with attribute [`gcc_struct`](https://gcc.gnu.org/onlinedocs/gcc/x86-Variable-Attributes.html), which will limit the above to just this struct, which is probably fine if this is an internal struct

lperron:
We have implemented the unsigned int version. 

I will push it on main soon.

PR #4312, opened 2024-07-15
Implement solver_infinity for XPRESS and make it available in Java

pet-mit:
- Implement solver_infinity for XPRESS and make it available in Java
- Add a non-regression test for this reacent bug fix: https://github.com/google/or-tools/pull/4214


pet-mit:
cleaned up the commits here : https://github.com/google/or-tools/pull/4313

PR #4339, opened 2024-08-03
Add `absl/log/vlog_is_on.h` `#include`s

ZhongRuoyu:
Abseil 20240722.0 has made it mandatory to explicitly `#include` `vlog_is_on.h` in order to use `VLOG_IS_ON` [^1]. Add those `#include`s so that the code builds with the latest Abseil.

[^1]: https://github.com/abseil/abseil-cpp/releases/tag/20240722.0

### DIFF
--- a/ortools/bop/bop_portfolio.cc
+++ b/ortools/bop/bop_portfolio.cc
@@ -24,6 +24,7 @@
 #include <vector>
 
 #include "absl/log/check.h"
+#include "absl/log/vlog_is_on.h"
 #include "absl/strings/str_format.h"
 #include "absl/strings/string_view.h"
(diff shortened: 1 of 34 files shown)


Mizux:
not sure about this, I would like to keep the code inline with the internal one.

Need to check if we get vlog_is_on from https://github.com/google/or-tools/blob/stable/ortools/base/logging.h 
i.e. add it there

Mizux:
Fixed by:

* https://github.com/google/or-tools/commit/9509cabec12d930d752add2471fa20617c1948ff
* https://github.com/google/or-tools/commit/05bfe988e498063942de2a253692828df57d4b5a

ZhongRuoyu:
Thanks @Mizux!

PR #4362, opened 2024-08-28
Make ortools/sat/go work with bazel

luluz66:
I made an attempt to make the go library build with bazel. 
- WORKSPACE: 
  - bumped rules_proto to 6.0.0
  - added dependencies for go and gazelle
- ortools/sat/go:
  - created a directory cpmodel and moved the go and cpp code under the directory; so that it followed golang convention where the package name and directory name are the same
  - go code: change imports, int type and proto type to make it compile
  - c++ code: I have to copy the code for `memdup`; because it's in absl/strings/internal/memutil.h and it's not visible.

Test:
I created a [test repo](https://github.com/luluz66/solver-test) that depends on this library; and I can run the binary with
```
bazel run :goexample
```


lperron:
I merged it.

Thanks!!!

it would be nice if you could update your repo to use the or-tools directly.

PR #4401, opened 2024-10-06
[WIP] Add the horizontal elastic timetable overload checking rule to the cumulative constraint. 

rhartert:
This PR adds an implementation of the Horizontal Elastic TimeTable Overload Checker presented in [Kameugne et al. 2024](https://scholar.google.fr/citations?view_op=view_citation&hl=fr&user=agmFWy8AAAAJ&sortby=pubdate&citation_for_view=agmFWy8AAAAJ:roLk4NBRz8UC) to the Cumulative Constraint.

The propagator is not activated by default and can be enabled by setting the `SatParameters.use_horizontal_overload_checking_in_cumulative` to `true`. 

## For Reviewers 

The PR is not ready for merge. Rather, it is meant as a "snapshot" to start discussing the implementation and whether the propagator is a net-positive for OR-Tools.

Ultimately, I plan to keep working on this PR to implement the "IQHE Edge-finder" rule (from the same paper) which dominates the TimeTable Edge-finder filter rule in terms of propagation. That being said, it is not clear whether this will be a net-positive in an LCG solver. The reason is that the explanations tend to be big and not easy to generalize.

### DIFF
--- a/ortools/sat/timetable_horizontal_edgefinding.h
+++ b/ortools/sat/timetable_horizontal_edgefinding.h
@@ -62,12 +62,12 @@ class HorizontallyElasticOverloadChecker : public PropagatorInterface {
   // profile.
   struct ProfileEvent {
     /* const */ int task_id;
-    /* const */ IntegerValue time;
-    /* const */ IntegerValue height;
     /* const */ ProfileEventType event_type;
+    IntegerValue time;
+    IntegerValue height;
 
-    ProfileEvent(int task_id, IntegerValue time, IntegerValue height, ProfileEventType event_type)
-        : task_id(task_id), time(time), height(height), event_type(event_type) {}
+    ProfileEvent(int task_id, ProfileEventType event_type, IntegerValue time, IntegerValue height)
+        : task_id(task_id), event_type(event_type), time(time), height(height) {}
 
     bool operator<(const ProfileEvent& other) const {
       return time < other.time;


ortools/sat/timetable_horizontal_edgefinding.cc
```diff
@@ -0,0 +1,290 @@
...
+                                   demands_->DemandMin(i),
+                                   ProfileEventType::kEndMin);
+      profile_events_.emplace_back(i, helper_->EndMax(i),
+                                   demands_->DemandMin(i),
+                                   ProfileEventType::kEndMax);
+    }
+  }
+  profile_events_.emplace_back(-1, kMaxIntegerValue, IntegerValue(0),
+                               ProfileEventType::kSentinel);
+  IncrementalSort(profile_events_.begin(), profile_events_.end());
```
vitor1001, line 89: I think that if you are not updating `profile_events_` incrementally there is no point on sorting it with this method. I think plain `std::sort` would be cleaner.

ortools/sat/timetable_horizontal_edgefinding.cc
```diff
@@ -0,0 +1,290 @@
...
+    if (event.event_type != ProfileEventType::kEndMax ||
+        event.time <= window_end) {
+      continue;
+    }
+    window_end = event.time;
+
+    const IntegerValue capacity = integer_trail_->UpperBound(capacity_);
+    if (window_end < ScheduleTasks(window_end, capacity)) {
+      AddScheduleTaskReason(window_end);
+      return helper_->ReportConflict();
```
vitor1001, line 104: If end_min is lower than the result of ScheduleTasks, can't you push a tighter end_min?

ortools/sat/timetable_horizontal_edgefinding.cc
```diff
@@ -0,0 +1,290 @@
...
+    IntegerValue delta_req = 0;
+    while (profile_events_[next_event].time == time) {
+      const ProfileEvent event = profile_events_[next_event];
+      switch (task_types_[event.task_id]) {
+        case TaskType::kIgnore:
+          break;  // drop the event
+        case TaskType::kFull:
+          switch (event.event_type) {
+            case ProfileEventType::kStartMin:
+              delta_max += event.height;
```
vitor1001, line 171: nit: I think changing demand_req and demand_max directly without the deltas might be simpler to understand.

ortools/sat/timetable_horizontal_edgefinding.cc
```diff
@@ -0,0 +1,290 @@
...
+          break;
+      }
+      next_event++;
+    }
+
+    // Should always be safe thanks to the sentinel.
+    DCHECK_LT(next_event, profile_events_.size());
+
+    IntegerValue next_time = profile_events_[next_event].time;
+    const IntegerValue length = next_time - time;
```
vitor1001, line 206: nit: please move this definition closer to where it is used.

ortools/sat/timetable_horizontal_edgefinding.cc
```diff
@@ -0,0 +1,290 @@
...
+
+    // Indicate whether we're using some capacity at this time point. This is
+    // used to decide later on how to update new_window_end.
+    const IntegerValue capa_used = std::min(demand_req + overload, true_capa);
+
+    // Amount of resource available to potentially place some overload from
+    // previous time points.
+    const IntegerValue overload_delta = demand_req - true_capa;
+
+    if (overload_delta > 0) {  // adding overload
```
vitor1001, line 226: nit: I think this logic would be easier to read if the variable naming made clear what is the overload/removable/used for a time unit and what is the equivalent for the time window.

vitor1001 (commented):
Thanks a lot for the PR!

I had a quick look at the CL and added some comments. I'm a new member of the team, so don't be surprised if I misunderstood something during the review.

ortools/sat/timetable_horizontal_edgefinding.cc
rhartert: You're right, changed the code to truly build the profile incrementally by updating the events rather than building it from scratch each time. 

I don't think this drives much benefits yet but it opens the door for further incremental optimization using reversibles similarly to what is done in time-table.

ortools/sat/timetable_horizontal_edgefinding.cc
rhartert: Agreed, done 👍

ortools/sat/timetable_horizontal_edgefinding.cc
rhartert: Given that `length` is only used once, I've inlined it directly in the overload adjustment. IMO this does not hurt the code's readability as the computation is small and pretty clear.

ortools/sat/timetable_horizontal_edgefinding.cc
rhartert: I wasn't super happy with these names either; what do you think about the new ones? I've also slightly simplified the logic as there's no need for the `std::min` call.

ortools/sat/timetable_horizontal_edgefinding.cc
rhartert: Good question, I guess we could use `ScheduleTasks(window_end, capacity) - 1` but I would have to think a little more about it. 

That being said, I don't think propagator as it stands (compared to its extensions that update the start/end vars) is worth adding to OR-tools. I'd rather prefer to optimize the explanation of these versions.

rhartert:
> Thanks a lot for the PR!
> 
> I had a quick look at the CL and added some comments. I'm a new member of the team, so don't be surprised if I misunderstood something during the review.

Hi @vitor1001, thank you for the review and the comments! I tried to address most of them, PTAL.

As mentioned in the PR description, I'd like to keep working on this PR and update the propagator to actually filter the start/end variables. Though, I would need a way to evaluate that the code is actually driving improvements (in terms of #nodes first, then in terms of speed). Is there an c++ example I could use to compare the performance of different versions of the cumulative on a benchmark that matters to the team by any chance?

lperron:
[rcpsp.tar.gz](https://github.com/user-attachments/files/17475553/rcpsp.tar.gz)

Here is a set of cp-sat protobufs for rcpsp models (single and multi mode)

PR #4417, opened 2024-10-21
HiGHS updates 

galabovaa:
- Highs version bump including fix of failing mathopt test
- methods added to highs_proto_solver for model, variable and constraint names and hints

PR #4418, opened 2024-10-21
remove skip test mathopt highs

galabovaa:
follow up of:
* #4417


galabovaa:
@lperron

galabovaa:
I will double check that it passes for me locally!

galabovaa:
I get a failure if I comment out 

https://github.com/google/or-tools/blob/0cf8ff8702d11a22cbad3698e79d644f2eb9f96c/ortools/math_opt/solver_tests/status_tests.cc#L376-L378

I will need to look into it further, so let's leave it for the next PR.

PR #4428, opened 2024-10-30
Add support for Bzlmod

mering:
Pending on https://github.com/bazelbuild/bazel-central-registry/pull/3066


mering:
Thanks to @mmorel-35 for many BCR uploads of dependencies!

PR #4447, opened 2024-11-21
Handle GLPK library naming on windows

jamiecook:
libglpk.a on linux, GLPK.lib on windows.

### DIFF
--- a/cmake/bundle-install.cmake.in
+++ b/cmake/bundle-install.cmake.in
@@ -34,7 +34,7 @@ if (@BUILD_SCIP@)
   set(DEPS ${DEPS} libscip.lib)
 endif()
 if (@BUILD_GLPK@)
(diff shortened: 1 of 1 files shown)


Mizux:
AFAIK this file is for windows only but:
0. I didn't bump it since we use dynamic library (protobuf abseil-cpp) -> May need some extra changes
1. Why talking about Linux, this file is only a post process for windows
2. In that case why using `*glpk.lib` instead of `glpk.lib` i.e. do you have some use case where the "*" is necessary ?

BTW thx for this PR we definitively need to take a look at this file before next release

jamiecook:
I missed that this was windows only, but that is obviously true with the
.lib extension.

When compiling on windows (so I can use glpk) I get a can't find file
error, libglpk.lib doesn't exist but GLPK.lib does.

- Jamie
http://jamiecook.wordpress.com

PR #4452, opened 2024-11-27
Support for MOSEK, linear, conic and mip

ulfworsoe:
As per our correspondence, here is  a PR for MOSEK support. I have not signed the contributor license yet - let me know if that will be necessary - otherwise, feel free to copy anything in this PR.

It should support most major features: Linear interface and math_opt, conic constraints, quadratic terms, integer variables, indicator constraints, deleting items (we do not actually remove constraints, only disable them).


rma350:
Looks great. At a high level, I think this is pretty similar to what we had done internally (wrapped the C API with ABSL/C++ bindings first).

I see that you clear variable/constraints instead of deleting them in your C++ wrapper. Your API does support deletion yes? I understand that in some settings, clearing is faster than deletion, but I guess our perspective is that the user can clear themselves if they want clear. If they really need deletion (e.g. they are managing a very large number of cuts/columns), then we want to offer it to them. Also, with your deletion API, do variables have stable ids, or indices, and all the indices change on each deletion?

 A big challenge that is not yet complete is to pass the integration tests, e.g. have a file like this:

https://github.com/google/or-tools/blob/stable/ortools/math_opt/solvers/gscip_solver_test.cc

(The gurobi version of this file appears not be exported to the open source project? Not sure why.) This always ends up being harder than it looks, due to a combination of the tests being a bit fragile and the solvers all being a little different (we try to run the same tests for all the solvers, with some solver specific configuraiton).

Specifically re MPSolver, inside of Google we are basically done adding features here, but we are continuing to work on the MPModelProto path e.g. here:

https://github.com/google/or-tools/tree/stable/ortools/linear_solver/proto_solver
https://github.com/google/or-tools/blob/stable/ortools/linear_solver/solve_mp_model.h

(but nothing conic). You would need to check with lperron, maybe we don't even want to merge any more mpsolver changes that don't just delegate to solving with the proto.

The people working the Mosek stuff are US based, so we will probably look more closely after thanksgiving.

cmake/FindMOSEK.cmake
```diff
@@ -0,0 +1,199 @@
...
+# Option 1. The user can specify when running CMake where the MOSEK platform directory is located, e.g.
+#     -DMOSEK_PLATFORM_DIR=$HOME/mosek/10.2/tools/platform/linux64x86/ 
+#   in which case no search is performed.
+# Option 2. The user can specify MOSEK_ROOT when running cmake. MOSEK_ROOT is
+#   the directory where the root mosek/ tree is located.
+# Option 3. Automatic search. We will then attempt to search in the default
+#   locations, and if that fails, assume it is installed in a system location.
+# For option 2 and 3, the newest MOSEK version will be chosen if more are available.
+
+if(MOSEK_PLATFORM_DIR)
```
Mizux, line 135: I would use `MOSEK_ROOT_DIR` to follow convention

From `CMake` head repository:
```sh
$ git grep -n "ROOT_DIR\b" | wc -l
209
$ git grep -n "PLATFORM_DIR\b" | wc -l
0
```

PR #4481, opened 2024-12-26
Update change of C API function name PyEval to PyObject

dodotronix:
Compilation of the or-tools fails, because the PyEval_CallFunction was replaced with PyObject_CallFunction.


lperron:
Already done on main and v99bugfix.

PR #4519, opened 2025-01-25
Dropped orders should not come with dropped reload stations in cvrp_r…

arnabanimesh:
…eload.py example output


Mizux (approved):
LGTM good catch !

Mizux:
Will merge it on Monday, want to perform the same fix on the C++, Java and .Net samples if any...

arnabanimesh:
Updated the Python notebook as well

Mizux:
> Updated the Python notebook as well
Thx for notebook update, please note usually we regenerate them just before a release using this script:
https://github.com/google/or-tools/blob/main/tools/generate_all_notebooks.sh

TODO(mizux): need to modify it to add your metadata update

Mizux:
Script fixed in 6eb08021632bc7652c30120759d02c57f6821d51

PR #4596, opened 2025-03-18
V9.12 Fix the lnk1189 problem

xingfutuhuangdi:
cmake 3.31.3
vs2022
window10

编译debug版本总是报lnk1189的问题，现在修复了。
Compiling debug version always quote lnk1189 issue now fixed.


CMakeLists.txt
```diff
@@ -443,22 +454,22 @@ include(dotnet)
...
- graph
- glop
- constraint_solver
- linear_solver
- ${MATH_OPT_DIR}
- ${PDLP_DIR}
- sat)
-  add_subdirectory(ortools/${SAMPLES}/samples)
-endforeach()
+# foreach(SAMPLES IN ITEMS
```
lperron, line 457: why this (and below) ?

CMakeLists.txt
```diff
@@ -30,6 +30,17 @@ message(STATUS "${PROJECT_NAME} version: ${PROJECT_VERSION}")
 
 if(MSVC)
   set(CMAKE_CXX_STANDARD 20)
+  # 设置源代码和执行字符集为 UTF-8（MSVC 2015 及以上支持 /utf-8）
```
Mizux, line 33: please use english in comment
why MSVC 2015 ? we only support 2022...

CMakeLists.txt
```diff
@@ -30,6 +30,17 @@ message(STATUS "${PROJECT_NAME} version: ${PROJECT_VERSION}")
...
+  message(STATE "set utf-8=======")
```
Mizux, line 34: there is no `STATE` and this log seems a debug log/dev to suppress imho

CMakeLists.txt
```diff
@@ -30,6 +30,17 @@ message(STATUS "${PROJECT_NAME} version: ${PROJECT_VERSION}")
...
+  add_compile_options("$<$<C_COMPILER_ID:MSVC>:/utf-8>")
+  add_compile_options("$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
+  # message(STATE "set bigobj=======")
+  add_compile_options(/bigobj)
+  SET(CMAKE_CXX_FLAGS_DEBUG "/Od2")
```
Mizux, line 39: why ? I prefer to be conservative and keep the cmake default debug configuration

CMakeLists.txt
```diff
@@ -30,6 +30,17 @@ message(STATUS "${PROJECT_NAME} version: ${PROJECT_VERSION}")
...
+
+  # 设置堆栈保留大小为16MB（单位：字节）
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /STACK:104857600")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} /STACK:104857600")
```
Mizux, line 43: please explain

Mizux:
this seems to have been IA generated....

CMakeLists.txt
xingfutuhuangdi: sorry，this is my fault

CMakeLists.txt
xingfutuhuangdi: My releas compilation caused errors in lnk1189, and I found two solutions on the Internet, so I added both of them. The compilation time was too long, so I added both solutions.

xingfutuhuangdi: about '/bigobj'



https://learn.microsoft.com/en-us/cpp/build/reference/bigobj-increase-number-of-sections-in-dot-obj-file?view=msvc-170

xingfutuhuangdi:
I'm not very good at English, so I have to use a translation software.
I pulled the v9.12 code for compilation, and I needed a debug version, but it failed.
I had to look for ways on the Internet. I had to try them one by one, and it took about three or four hours to compile them.
There are some things I don't know about, it just work. Thank you for your efforts, and I hope these codes won't cause you any trouble.

CMakeLists.txt
xingfutuhuangdi: Set the stack retention size.

CMakeLists.txt
xingfutuhuangdi: Sorry, it was my fault. I forgot to remove the comment. Very Very Sorry.

xingfutuhuangdi: Sorry, it was my fault. I forgot to remove the comment. Very Very Sorry.

CMakeLists.txt
Mizux: IIRC we already enable `/bigobj`...
https://github.com/google/or-tools/blob/35c27ab31f8a093cdc792144c2d3deb55ed11521/cmake/cpp.cmake#L102-L105

Mizux:
on `main` branch, ortools has been splitted into multiple libraries to avoid the maximum symbols error aka this error should be fixed on `main`

PR #4601, opened 2025-03-26
Misleading method name fix in SetCoverModel

c4v4:
Hello!  

I'm currently working on a larger PR to (hopefully) introduce a subgradient-based heuristic for the Set Covering problem. In the process, I noticed a couple of minor issues that might be worth addressing:  

### Rename `ReserveNumSubsets` --> `ResizeNumSubsets`  
The name `ReserveNumSubsets` suggests (by analogy with `std::vector::reserve`) that it only allocates memory without creating new subsets. However, it actually resizes the model, introducing the necessary empty subsets.

Moreover, since the current interface allows models to be built subset by subset (with `AddElementToLastSubset` behaving similarly to `std::vector::push_back`), this further increases the risk of misusing the function.

Additionally, `ReserveNumElementsInSubset` (defined right after it) does follow the expected behavior of `std::vector::reserve`, meaning it allocates memory but does not create new elements. 

To remove this inconsistency, this PR renames `ReserveNumSubsets` to `ResizeNumSubsets`, aligning it with `std::vector::resize` to better reflect its behavior.

### Nit: `ElementIndex`  -->  `RowEntryIndex`  
Since the codebase defines `RowEntryIndex` as the custom type for row entry indices, this PR also replaces `ElementIndex` with `RowEntryIndex`  on the `ReserveNumElementsInSubset` overload.  

Let me know if you have any feedback!

### DIFF
--- a/ortools/algorithms/python/set_cover.cc
+++ b/ortools/algorithms/python/set_cover.cc
@@ -221,9 +221,9 @@ PYBIND11_MODULE(set_cover, m) {
       .def("sort_elements_in_subsets", &SetCoverModel::SortElementsInSubsets)
       .def("compute_feasibility", &SetCoverModel::ComputeFeasibility)
       .def(
-          "reserve_num_subsets",
+          "resize_num_subsets",
           [](SetCoverModel& model, BaseInt num_subsets) {
-            model.ReserveNumSubsets(num_subsets);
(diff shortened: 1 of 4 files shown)


lperron:
I pushed the proposed changes.

I will close this PR.

Thanks.

Note: we are looking at improving heuristics. You should reach out to collaborate with out team.

PR #4618, opened 2025-04-02
add IDisposable to CPSolver

ChangeTheCode:
## Description
This PR adds the `IDisposable` interface to the `CPSolver` class to make the existing `Dispose` implementation more explicit and clear to users of the class.

The `CPSolver` class already had a `Dispose` method implemented to clean up any unmanaged resources used by the class. By explicitly implementing the `IDisposable` interface, we make it clear that the class follows the standard .NET dispose pattern and can be properly disposed of to release those resources.


ChangeTheCode:
Provided CLA information

ChangeTheCode:
need to adapt changes to the newest implementation

lperron:
can you target the main branch and not the stable one ?

ortools/sat/csharp/CpSolver.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2010-2025 Google LLC
+// Copyright 2010-2024 Google LLC
```
lperron, line 1: revert that one :-)

ortools/sat/csharp/CpSolver.cs
```diff
@@ -11,12 +11,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
-using System.Collections.Generic;
 using System.Runtime.CompilerServices;
+using Google.OrTools.Sat;
 
-namespace Google.OrTools.Sat
```
Mizux, line 18: Why did you remove the namespace ?
i.e. `CpSolver` should still be inside the `Google.OrTools.Sat` IMHO (and the new `using` directive won't make it)

ortools/sat/csharp/CpSolver.cs
```diff
@@ -26,238 +23,175 @@ namespace Google.OrTools.Sat
  * variables in the best solution, as well as general statistics of the search.
  * </remarks>
  */
-public class CpSolver
+public sealed class CpSolver : IDisposable
 {
-    /** <summary>Solves the given model, and returns the solve status.</summary> */
-    public CpSolverStatus Solve(CpModel model, SolutionCallback cb = null)
+    private LogCallback? _log_callback;
```
Mizux, line 28: Google **C++** coding convention use `_` as suffix for private member why did you choose to use `_` as prefix ?
Is it backed by a .Net naming convention (and tooling) ? 

ref: https://google.github.io/styleguide/cppguide.html#Variable_Names

ortools/sat/csharp/CpSolver.cs
```diff
@@ -11,12 +11,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
```
Mizux, line 14: `IDisposable` seems to comes from `System` why did you remove it ?
https://learn.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-8.0

ortools/sat/csharp/CpSolver.cs
ChangeTheCode: prefix is a standard of c#. I normally use analyzer to enforce static code analyses. But it is fine. I will follow your convention to have an uniformed way 👌

ChangeTheCode:
> can you target the main branch and not the stable one ?
I changed the target from `stable` to `main`

ortools/sat/csharp/CpSolver.cs
Mizux: @ChangeTheCode AFAIK Google do not provide a naming convention for .Net (i.e. not an official language here) so in this case we prefer to stick with "default language/microsoft" naming convention so we can leverage ecosystem tooling -> I would prefer to keep the `_` prefix if it reduce the friction...

Mizux: > [...] I normally use analyzer to enforce static code analyses.

May I have the name of your "analyzer tools" i.e. how I can reproduce the errors you got "on my machine (TM)"

ChangeTheCode: I can create therefore, a separate PR. 
Lets focus on the disposable part

lperron:
I will merge it manually.

I fixed the double dispose () method.

00


Le mer. 9 avr. 2025, 10:07, ChangeTheCode ***@***.***> a
écrit :

lperron:
Merged manually.

ChangeTheCode:
@lperron  do you new when this will be released in a new version ?

lperron:
as soon as math_opt python compiles.
Tentative early next week. It will most likely be delayed.
00



Le mer. 9 avr. 2025 à 15:57, ChangeTheCode ***@***.***> a
écrit :

lperron:
In the meantime, you can build from source using the v99bugfix branch. It
contains the new CpSolver.cs
00

PR #4669, opened 2025-06-04
Use a fixed order for MPSolver / XPRESS's constraint coefficients

flomnes:
Using `absl::flat_hash_map<const MPVariable*, double> coefficients_` from `MPSolver::MPConstraint` yields a different order for coefficients provided the `XPRSaddrows` method. This leads to non-reproducibility in cases where problems have non-unique solutions.

Instead we use our own map with a fixed order (ie not dependent on unpredictable pointer addresses).

@lperron @Mizux If it's OK for you, can this be shipped into v9.13 ?


flomnes:
- Use `absl::btree_map` for better performance
- Fix unit tests by improving synchronization of `fixedOrderCoefficientsPerConstraint`
   - Calls to `XpressInterface::ClearConstraint` had no effect
   - Multiples calls to `XpressInterface::SetCoefficient` were not taken into account properly

PR #4685, opened 2025-06-16
Fix bz2.dll install path for windows-cpp archive

flomnes:
In Windows, bz2.dll is incorrectly placed in install/lib, instead of install/bin. This causes "missing DLL error at runtime" kind of issues for binaries linking dynamically to OR-Tools 9.13 through CMake.

This issue was [reported in 2022](https://gitlab.com/bzip2/bzip2/-/issues/45) in bzip2's official repository, but it was ignored by bzip2's maintainers.

bzip2 was introduced as a dependency of OR-Tools from v9.13, which is why we have this problem right now.

Example of this [or-tools_x64_VisualStudio2022_cpp_v9.13.4784.zip ](https://github.com/google/or-tools/releases/download/v9.13/or-tools_x64_VisualStudio2022_cpp_v9.13.4784.zip)

image

image

### DIFF
--- a/patches/bzip2.patch
+++ b/patches/bzip2.patch
@@ -1,5 +1,5 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
-index c4b0b6e..30f7652 100644
+index c4b0b6e..ee39341 100644
 --- a/CMakeLists.txt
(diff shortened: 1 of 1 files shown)


Mizux (approved):
LGTM

Issue #4345, opened 2024-08-13
Solution hint is not passed to the HiGHS solver using mathopt


pavlomuts:
**What version of OR-Tools and what language are you using?**
Version: 9.10
Language: Python

**Which solver are you using (e.g. CP-SAT, Routing Solver, GLOP, BOP, Gurobi)**
HiGHS, SCIP

**What operating system (Linux, Windows, ...) and version?**
Ubuntu 22.04

**What did you do?**
Using the following simple model I am trying to pass to the solver a solution hint in order to warm start it.
```python
from ortools.math_opt.python import mathopt, model_parameters

model = mathopt.Model()
x = model.add_integer_variable(name="x")
y = model.add_integer_variable(name="y")

model.add_linear_constraint(2 * x + 2 * y >= 7)
model.add_linear_constraint(3 * x + 4 * y >= 12)
model.add_linear_constraint(2 * x + y >= 6)

model.minimize_linear_objective(8 * x + 10 * y)

params = mathopt.SolveParameters()
params.enable_output = True

solution_hint = model_parameters.SolutionHint(variable_values={x: 1, y: 25})
mathopt.solve(
    model,
    mathopt.SolverType.HIGHS,
    params=params,
    model_params=model_parameters.ModelSolveParameters(solution_hints=[solution_hint]),
)
```

The output below clearly indicates that the solution hint was not taken into account, as first `BestSol` is inf:
```
Coefficient ranges:
  Matrix [1e+00, 4e+00]
  Cost   [8e+00, 1e+01]
  Bound  [0e+00, 0e+00]
  RHS    [6e+00, 1e+01]
Presolving model
3 rows, 2 cols, 6 nonzeros  0s
3 rows, 2 cols, 6 nonzeros  0s
Objective function is integral with scale 0.5

Solving MIP model with:
   3 rows
   2 cols (0 binary, 2 integer, 0 implied int., 0 continuous)
   6 nonzeros

        Nodes      |    B&B Tree     |            Objective Bounds              |  Dynamic Constraints |       Work      
     Proc. InQueue |  Leaves   Expl. | BestBound       BestSol              Gap |   Cuts   InLp Confl. | LpIters     Time

         0       0         0   0.00%   -inf            inf                  inf        0      0      0         0     0.0s
 T       0       0         0   0.00%   -inf            32                 Large        0      0      0         3     0.0s

Solving report
  Status            Optimal
  Primal bound      32
  Dual bound        32
[...]

PR #4737, opened 2025-07-22
Pass solution hint to HiGHS from mathopt

pavlomuts:
Issue:
* #4345 
@rma350

### DIFF
--- a/ortools/math_opt/solvers/highs_solver_test.cc
+++ b/ortools/math_opt/solvers/highs_solver_test.cc
@@ -164,9 +164,18 @@ INSTANTIATE_TEST_SUITE_P(HighsLpModelSolveParametersTest,
                              /*supports_duals=*/true,
                              /*supports_primal_only_warm_starts=*/true)));
 
-// Highs::setSolution is implemented, but it only accepts complete solutions.
-// The test below generates partial solutuions, so we skip it.
-GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(MipSolutionHintTest);
+SolutionHintTestParams MakeHighsSolutionHintParams() {
+  SolveParameters solve_params;
+  solve_params.presolve = Emphasis::kOff;
+  (*solve_params.highs.mutable_int_options())["mip_max_nodes"] = 0;
+  std::string hint_message_regex =
+      "Attempting to find feasible solution by "
+      "solving MIP for user-supplied values of";
+  return SolutionHintTestParams(SolverType::kHighs, solve_params, std::nullopt,
+                                hint_message_regex);
+}
+INSTANTIATE_TEST_SUITE_P(HighsSolutionHintTest, MipSolutionHintTest,
+                         Values(MakeHighsSolutionHintParams()));
 
 // HiGHS does not support branching priority.
 GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(BranchPrioritiesTest);
--- a/ortools/math_opt/solvers/highs_solver_test.cc
+++ b/ortools/math_opt/solvers/highs_solver_test.cc
@@ -164,8 +164,8 @@ INSTANTIATE_TEST_SUITE_P(HighsLpModelSolveParametersTest,
                              /*supports_duals=*/true,
                              /*supports_primal_only_warm_starts=*/false)));
 
-// MIP hint appears to be supported by Highs::setSolution, this is not yet
-// implemented.
+// Highs::setSolution is implemented, but it only accepts complete solutions.
+// The test below generates partial solutuions, so we skip it.
 GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(MipSolutionHintTest);
 
 // HiGHS does not support branching priority.
--- a/ortools/math_opt/solvers/highs_solver_test.cc
+++ b/ortools/math_opt/solvers/highs_solver_test.cc
@@ -162,7 +162,7 @@ INSTANTIATE_TEST_SUITE_P(HighsLpModelSolveParametersTest,
                          Values(LpModelSolveParametersTestParameters(
                              SolverType::kHighs, /*exact_zeros=*/true,
                              /*supports_duals=*/true,
-                             /*supports_primal_only_warm_starts=*/false)));
+                             /*supports_primal_only_warm_starts=*/true)));
 
 // Highs::setSolution is implemented, but it only accepts complete solutions.
 // The test below generates partial solutuions, so we skip it.
--- a/ortools/math_opt/solvers/highs_solver.cc
+++ b/ortools/math_opt/solvers/highs_solver.cc
@@ -925,6 +925,22 @@ absl::StatusOr<SolveResultProto> HighsSolver::Solve(
     return absl::OkStatus();
   };
 
+  if (model_parameters.solution_hints_size() > 0) {
+    // Take the first solution hint and set the solution.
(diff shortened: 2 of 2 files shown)


ortools/math_opt/solvers/highs_solver_test.cc
```diff
@@ -164,8 +164,8 @@ INSTANTIATE_TEST_SUITE_P(HighsLpModelSolveParametersTest,
                              /*supports_duals=*/true,
                              /*supports_primal_only_warm_starts=*/false)));
 
-// MIP hint appears to be supported by Highs::setSolution, this is not yet
-// implemented.
+// Highs::setSolution is implemented, but it only accepts complete solutions.
```
rma350, line 167: We cannot submit code without tests. As commented above, perhaps you can get the partial hint working.

If not, then you need to fork this test:

https://github.com/google/or-tools/blob/stable/ortools/math_opt/solver_tests/ip_model_solve_parameters_tests.cc#L101

in the same file to add a version of the test that has a complete hint.

A fallback would be to add the fork of this test directly to highs_solver_test.cc that does this, if you cannot get it working for every solver.

Hopefully, partial hints are actually supported.

rma350 (changes requested):

rma350:
This change looks good to me. Not sure if Mizux has any further comments. There appears to be some issue with the CLA.

rma350 (approved):

Mizux:
CLA is mandatory for us to be able to merge the commit...

pavlomuts:
Thanks, guys. I am trying to figure out how should I proceed with CLA, will get back later as soon as I have the clarity.

pavlomuts:
Signed the CLA, thank you all!

pavlomuts:
@Mizux @rma350 Could you please advice what should I do to merge this PR? The CLA is signed.

rma350:
@Mizux is in vacation this week. I think it is good to go, let me ping lperron.

temph2020:
@pavlomuts mathopt is slow at generating constraints. So I still stick to pywraplp, could you modify it to Pass solution hint to HiGHS from pywraplp?

rma350:
FWIW, here is a measurement I made on time to build and solve ~1 year ago.

Time to build and the model, solve, and consume the solution for a random knapsack problem with 100K items with various google3 python MIP APIs/strategies:

API/Strategy                                                                 Time (s)
MathOpt Python w/ Linear Expressions                       4.82
MathOpt Python w/o Linear Expressions                     3.29
MathOpt Python with NumPy                                       1.78    (not sumbitted, more optimizations possible)
MPSolver w/o Linear Expressions                                2.73
ModelBuilder with WeightedSum                                  2.27

Actual Gurobi time for this model was ~0.7 second, so we are in the (rare) regime where model building time dominates.

Since then, lperron has done some more work on model builder to (1) directly integrate model builder with pandas and (2) move the linear expression represention to C++ (there are some tradeoffs here, pybind11 is also slow and this is more corssings, but I think it was a net win on whatever he was measuring), so if you want speed on submitted code, this is the way to go.

(Internally, so go/mathopt-numpy for details, the numpy part is exposed only up to elemental, it was never connected end-to-end, anyone adventerous can reach inside at their own risk.)

I would recommend against continuing with pywraplp/MPSolver, no one is working on this (no new features coming from google, no hope of a future extension of the API for better bulk operations.  If you avoid linear expressions, you can get very similar performance using mathopt (I think pybind11 as we use it is a little slower than SWIG).

Not 100% sure what your problem is, but for 95%+ of users (even in python) the time to build the model is much smaller than the time to solve it. The random knapsack exmaple above was chosen as a worst case example since the solution is somewhat trivial as compared to the size of the model. If your solve time is 20 minutes, does the model building time actually matter?

What is going on in the mathopt code re performance? The linear expression code is missing an optimization for avoiding lots of allocations when you need to compute an inner product that model builder has. If you are just summing a lot of things, we do have an optimization to avoid many allocations, you should use fast_sum instead of the built in sum function or repeated + operator. No matter what you do though, Pybind11 is going to be a bottleneck (200 nanoseconds per crossing) so really you need some kind of numpy represenation of your problem to avoid crossing the boundary for every nonzero in your problem.

There also just might be something wrong with your code (it is easy to accidentally get quadraitc behavior with linear expressions if you hold them wrong). If you have a minimal example you can post, I will try to take a quick look (no promises...).

temph2020:
＠[rma350](https://github.com/rma350)
[test2.py](https://github.com/user-attachments/files/27481253/test2.py)
pls see the attached. if you run solve_with_mathopt(data), time to generate constraints is 50s. if solve_model_builder(data), it is 15s. pls see the logs.

rma350:
I looked over your code and it seems fine, nothing stuck out. I think that this is roughly what I saw in the knapsack example (a little worse for mathopt, lperron has further optimized the linear expressions for model builder since I did my measurement).

I didn't execute the code, but it looks like the model is quite big. Can you actually sovle it quickly? If the model building time is large compared to solve time, this is where we try to optimize, but for users where solve is also slow we tend to worry less about it.

Regardless, the plan to do better was numpy support in the mathopt API, but that project is on pause for now. You can try just not using linear expressions and the code will go faster if this is important, but this is not a good user experience.

temph2020:
Actually, what we need most is the hint functionality of HiGHS. Since mathopt supports HiGHS hints, we initially planned to migrate to mathopt, but the constraint generation time is too long and unacceptable.
On the other hand, modelbuilder is fast at generating constraints, but it does not support the hint functionality for HiGHS. As a result, we are now caught in a dilemma.
The hint functionality is critical for us because we use a multi‑stage solving process: subsequent steps require the solution from the previous step as a hint; otherwise, finding an initial feasible solution is extremely slow. Therefore, is it possible to add HiGHS hint support to modelbuilder or pywraplp? [rma350](https://github.com/rma350)

lperron:
It is added on the main branch for modelbuilder

00

temph2020:
@lperron I tested main branch for modelbuilder for highs's hint just now, it didn't work.

temph2020:
in the main branch, in highs_proto_solver.cc, AI said:
 std::vector<HighsInt> hint_index(0, num_hints); //error, size is 0
  std::vector<double> hint_value(0, num_hints);  //error, size is 0 @rma350 @Mizux @lperron

temph2020:
by the way, in [test2.py](https://github.com/user-attachments/files/27481253/test2.py), if you set solve using gurobi, then in main branch or v9.15, no output.  in v9.14, it has output.

lperron:
Can you send a pr with the fix for highs against main ?

00

temph2020:
I'm not familiar with how to make the modifications. Here are the revisions( in highs_proto_solver.cc) suggested by Gemini, but my tests show they still don't work. I was wondering if someone with more experience could take a look and make further corrections.

    // Hints.
    int num_hints = model.solution_hint().var_index_size();
    if (num_hints > 0) {
      std::vector<HighsInt> hint_index;
      hint_index.reserve(num_hints);
      std::vector<double> hint_value;
      hint_value.reserve(num_hints);
      for (int i = 0; i < num_hints; ++i) {
        hint_index.push_back(model.solution_hint().var_index(i));
        hint_value.push_back(model.solution_hint().var_value(i));
      }
      HighsStatus status =
          highs.setSolution((HighsInt)num_hints, hint_index.data(),
                            hint_value.data());
      if (status == HighsStatus::kError) {
        response.set_status(MPSOLVER_MODEL_INVALID_SOLUTION_HINT);
        response.set_status_str("setSolution failed");
        return response;
      }
    }

PR #4756, opened 2025-08-21
backport stable to main

Mizux:


Mizux:
Seems already applied on main... :thinking:

PR #4813, opened 2025-09-21
fix binary command for rfz_% target

mathematikoi:


Mizux:
confirmed here:
https://github.com/google/or-tools/blob/debadf07dfd99ae730e0f9f147267b147d1af404/cmake/flatzinc.cmake#L132

Thx for the fix !

Issue #4133, opened 2024-03-15
OR-Tools v9.9 regression: incorrect negative transit variables (with modulo > capacity) handling in heterogeneous VRP


Yamp:
**What version of OR-Tools and what language are you using?**
Version: v9.9 only.
Language: Python

**Routing Solver**

**Macos 13.4**

I am using code like this to solve heterogeneous VRP with reloads. Like in this example
https://github.com/google/or-tools/blob/stable/ortools/constraint_solver/samples/cvrp_reload.py
But my problem has different capacities for different vehicles.

```py
  self.r.AddDimensionWithVehicleCapacity(
      evaluator_index=self.c[name],
      slack_max=int(self.p.max_c[index]),  # we are unloading max_cap but adding some slack
      vehicle_capacities=self.p.max_capacities[:, index].astype(int).tolist(),
      fix_start_cumul_to_zero=True,
      name="max_capacities" + str(index),
  )
```

Making negative capacities at reloads equal to the **biggest** possible capacity of my vehicle and then adding slack to landfills.
At version 9.8 it works, but at 9.9 it does not accept the initial solution and does not use landfills for vehicles with non-maximum capacities.

After some debugging, I realized that reload capacities say -100 are considered out of bounds for vehicles with maximum capacity say 80. So it cannot use reloads and ignores the initial solution. How to solve the same problem in the 9.9 version I don't know. I would consider this a bug.

cyril23:
We can confirm this bug still exists in OR-Tools 9.14.6206 (latest) and have identified the exact root cause through source code analysis.

## Root Cause

**Commit:** 847cfc9b93 (Dec 1, 2023)
**Function:** `FinalizeAllowedVehicles()` in `ortools/constraint_solver/routing.cc`
**Problem:** Incorrectly removes vehicles from `allowed_vehicles_` sets when dimensions use negative transit values

### Bug Mechanism

The function checks if `abs(transit) > capacity` for unary dimensions and removes vehicles that fail this check. This is incorrect for:

1. **Reload dimensions** (@Yamp 's original report): Negative values at landfills/depots
2. **CARGO_LOAD dimensions** (our use case): Negative constants used for cumulative load tracking

In our multi-class pickup/delivery scenario:
- CARGO_LOAD dimension uses `transit = -max(capacities)` for all nodes (e.g., -70,000)
- Smaller vehicles (capacity 50,000) fail the check: `abs(-70,000) > 50,000`
[...]

PR #4853, opened 2025-10-07
Fix issue #4133: FinalizeAllowedVehicles incorrectly excludes vehicles with negative transits

cyril23:
**Fixes:**
* #4133

## Problem

`FinalizeAllowedVehicles()` uses `abs(transit)` when checking if a vehicle can serve a node, which incorrectly excludes vehicles when dimensions use negative transit values for algorithmic purposes (reload dimensions, load tracking).

**Detailed analysis:** https://github.com/google/or-tools/issues/4133#issuecomment-3377648091

## Solution

Skip capacity checks for non-positive transits, since only positive transits represent actual cargo/capacity requirements:

```cpp
const int64_t transit = transit_evaluator(node);
if (transit <= 0 || transit <= capacity) continue;
```

## Changes

**File:** `ortools/routing/routing.cc` (after merge with main branch - file was moved from `ortools/constraint_solver/routing.cc`)

### 1. Pre-processing loop (~lines 1414-1421~ lines 1413-1420)
```cpp
for (int callback_index : dimension->class_evaluators_) {
  // Only consider positive transits for capacity checks.
  // Negative transits are used for reload/load tracking and should not
  // trigger vehicle exclusions based on capacity.
  const int64_t transit = UnaryTransitCallbackOrNull(callback_index)(node);
  if (transit > 0) {
    max_node_transit = std::max(max_node_transit, transit);
  }
}
```

### 2. Main capacity check (~lines 1444-1448~ lines 1443-1447)
```cpp
// Fix for issue #4133: Skip capacity check for negative transits.
// Negative transits are used for reload dimensions and load tracking,
// not actual cargo requirements.
const int64_t transit = transit_evaluator(node);
if (transit <= 0 || transit <= capacity) continue;
```

## Testing

Built and tested with the reproduction script from the issue comment:

**Before fix (v9.12.4544):**
```
Dropped customer stops: 13
Tours found: 1
```

**After fix:**
```
Dropped customer stops: 0
Tours found: 4
```

✅ All customer stops served, all vehicle classes utilized correctly

**Backward compatibility:** Only affects models using negative transits in unary dimensions. Standard capacity models are unaffected.

## edit: Merge Note                                                                                                                                                                                                                                                            

This PR has been rebased on the current `main` branch (v10.0), which includes a routing module reorganization where routing files were moved from `ortools/constraint_solver/` to `ortools/routing/`. The fix has been successfully applied to the new file location.


cyril23:
OK I've done the CLA and resolved the merge conflict. Ready for review.

lperron:
no, this correct is currently is correct.

cumul var needs to be 0 <= var <= max_value

if abs(transit) > max_value, the node can never be visited by that vehicle.

What the code does is change the variable ordering which leads to different solutions.

But the proposed fix is incorrect.

PR #4868, opened 2025-10-13
Fix issue #4133: Correct negative transit handling in FinalizeAllowedVehicles

### LINKED ISSUE #4133: OR-Tools v9.9 regression: incorrect negative transit variables (with modulo > capacity) handling in heterogeneous VRP

cyril23:
## Summary

This PR fixes a regression introduced in v9.9 (commit 847cfc9b93) where `FinalizeAllowedVehicles()` incorrectly excluded vehicles from nodes with negative unary transits, causing heterogeneous VRP problems with reload/load-tracking dimensions to produce suboptimal solutions with excessive dropped stops.

## The Bug

The optimization in `FinalizeAllowedVehicles()` checked:
```cpp
if (std::abs(transit) <= capacity) continue;
```

This is incorrect for negative transits. A vehicle with `capacity=50000` was excluded from nodes with `transit=-70000`, even when `slack_max=140000` was sufficient to absorb the negative transit.

## Why PR #4853 Was Incorrect

The previous fix attempted to skip all negative transits:
```cpp
if (transit <= 0 || transit <= capacity) continue;
```

As @lperron correctly noted, this breaks feasibility guarantees. A negative transit with `abs(transit) > slack_max` should genuinely exclude vehicles, but that fix allowed them through.

## The Correct Fix

Negative transits are feasible when the vehicle can absorb them via **combined capacity and slack**:

```cpp
const int64_t min_allowed_transit = CapSub(0, CapAdd(capacity, slack_max));
if (transit <= capacity && transit >= min_allowed_transit) continue;
```

This translates to:
- **Positive transits**: `transit ≤ capacity` (must fit in vehicle)
- **Negative transits**: `abs(transit) ≤ capacity + slack_max` (absorbed via capacity headroom + slack)

This is mathematically correct: a vehicle starting with `cumul = capacity` can decrease by up to `capacity + slack_max` before violating the lower bound (0).

## Regression Test

A regression test reproducing the exact bug is available: [test_issue_4133_regression.py](https://github.com/user-attachments/files/22887573/test_issue_4133_regression.py)

**Results with this fix:**
```
[PASS] No drops and multiple vehicle classes used (4 tours, vehicles [0, 1, 2, 3]) with OR-Tools 10.0.9999
```

**Expected results on v9.9-9.14 (broken versions):**
```
[FAIL] Regression detected: dropped customers [1, 8, 14, 16, 17, 18, 19, 20, 21, 22, 23, 28, 29] with OR-Tools 9.14.6206
```

## Impact

Fixes routing problems affected for 20+ months (v9.9.3963 through v9.14.6206):
- Heterogeneous VRP with reload dimensions
- Multi-class pickup/delivery with load-tracking dimensions
- Any scenario using negative unary transits for algorithmic purposes

Fixes #4133


StevenGay:
Thanks for your explanation and your code, I agree with the general modification, this was an actual bug.
We modified the code internally with a simpler change that should be more tight than your PR, it will be available shortly.

PR #4901, opened 2025-10-27
cmake: update highs to v1.12.0

galabovaa:
HiGHS v1.12.0 is out !
ref: https://github.com/ERGO-Code/HiGHS/releases/tag/v1.12.0


Mizux:
thx for the update need to check 2 points (RPATH management) the rest was trailing whitespace (and my vim conf auto trim upon saving)

galabovaa:
Thank you @Mizux, I added your patch code to HiGHS v1.12 to avoid the need for patches. Testing locally worked OK

PR #4908, opened 2025-11-05
cmake: system_deps build googletest

Mizux:
TLDR: have to built it to have string_view support

long story: currently you can built googletest with or without abseil-cpp, BUT if you built it without (like most distro do) then you won't have `std::string_view` (using `absl::string_view` alias) support in all googletest API...

in google3 we do use use absl::string_view with googletest so it is mandatory to have absl support in googletest when exporting or-tools tests...

### DIFF
--- a/cmake/docker/system_deps/Dockerfile
+++ b/cmake/docker/system_deps/Dockerfile
@@ -12,7 +12,7 @@ RUN pacman -Syu --noconfirm \
  coin-or-cbc coin-or-clp \
  glpk \
  scip \
- gtest benchmark
+ benchmark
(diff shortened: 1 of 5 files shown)


Mizux:
this PR fix compilation errors, but we still have some issue at runtime since archlinux already use/provide Protobuf v33.0 while `main` still use Protobuf v32.0 so the C++ is compiled against the system wide protobuf package (v33) while java (pom.xml) and python (setup.py) still require protobuf v32.

ref: https://archlinux.org/packages/extra/x86_64/protobuf/

PR #4888, opened 2025-10-21
Xpress use optimizer function

djunglas:
This pull requests adds support for a lot of MIP features for the Xpress solver in ortools:
1. Integer variables
2. Solution hints (interpreted as MIP starts)
3. Lazy constraints
4. Indicator constraints
5. SOS constraints (only on singleton variables, not on expressions)
6. Second order cone constraints (only on singleton variables, not on expressions)
7. Multi-objective
8. Quadratic constraints
9. First order methods
10. Message callbacks
11. Interrupters
12. Setting arbitrary Xpress specific parameters (via new `XpressParameters` class)
13. Support writing out models as they are optimized using `EXPORT_MODEL` parameter
14. Support for solve parameters (presolve, cut-off, thread count, time limit, node limit, absolute and relative gap tolerance, cuts, heuristics)
15. Support for changing Xpress controls, including some meta controls (`FORCE_POSTSOLVE` to force immediate post solve, `EXTRACT_MODEL` to extract the low-level model to disk for debugging, `STOP_AFTER_LP` to stop right after the root relaxation and earlier than `MAXNODE=0` would do)
16. Support for `InitArgs`: The now have an `xpress.extract_names` field that controls whether variable, constraint etc. names are extracted to the low level model

We also changed the LP part: we now always call XPRSoptimize(), which is a more modern way of using Xpress and letting Xpress decide what is the best algorithm to run.

Even with this pull request we still do not support the following things for the Xpress solver
- The callback
- Computing minimal infeasible subsystems
- The Update() function (for incremental solving)

We had to adjust some tests because it seemed they did not honor all the parameters passed into them. It is probably a good idea to double check these changes very carefully.

PR #4910, opened 2025-11-06
[MPSolver] Add lazy constraints support in XPRESS interface]

pet-mit:
ref: https://www.fico.com/fico-xpress-optimization/docs/latest/solver/optimizer/HTML/XPRSloaddelayedrows.html


lperron:
How does this play the other PR ?

pet-mit:
> How does this play the other PR ?

If you're asking about #4888, they are independent since this one is for MPSolver interface only. I made sure changes to xpress_environment h & cc are the same to avoid conflicts (my changes are included in the other changes, at the same lines)

PR #4961, opened 2025-12-20
Upgrade Java version to 21 across all build environments

gchatelet:
Todo list to migrate to JDK 21
* [x] ci: bump workflow
* [x] bazel: bump bazelrc
* [x] ci: bump bazel docker
* [x] cmake: bump cmake/java.cmake
* [x] ci: bump cmake docker
* [x] java: bump ortools/java/*pom.xml.in
* [x] release: bump tools/release dockerfiles
* [x] release: bump tools/docker
* [ ] release: update release machines
  * [ ] internal: bump internal machine setup documentation


Mizux:
few notes:
1. moving to openjdk21 in docker may not work as expected (sometime bazel doesn't catch it etc..)
-> you may need https://github.com/google/or-tools/commit/609b5db9124aac6527b2022c241f0548837a606c
since docker job are only trigger on push -> need to monitor once merge and apply this fix if needed

2. in java we set the jdk release in the pom.xml
https://github.com/google/or-tools/blob/af5487d1b1f46761dabac8ce7a59cf97e43d6025/ortools/java/pom-local.xml.in#L51-L55
maybe it's time to set it to 21 too or at least 11 
note: Actually java team failed the version support, and you need a third option on the cli `--release` IIRC so trying to build for 1.8 with 21. So while targeting 1.8 API you can have some byte code/implem relying on jdk11+ internal function...

3. we also need to bump some tools/release manylinux docker, as well as tools/docker/[images,test] to built against 21 to be consistent when building release artifact...

gchatelet:
For 1. this is now done at the [.bazelrc level](https://github.com/google/or-tools/pull/4961/files#diff-544556920c45b42cbfe40159b082ce8af6bd929e492d076769226265f215832f). So I think this is covered. Having a look at the rest.

gchatelet:
Should be fixed, let me know if I missed something.

PR #4962, opened 2025-12-21
Rename CPSolver to CpSolver

yoonghm:


yoonghm:
CPSolver should'be CpSolver

lperron:
Thanks

PR #4966, opened 2025-12-24
Amend #4961, bump Java 21.

gchatelet:


Mizux:
is there a reason to force the usage of temurin jdk instead of using the default jdk provides by ubunutu registry directly on linux and brew on macos (or the one in the runner directly) ?
also using the runner default env should improve interoperability and extend our test surface area for free...

ref:
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#java
https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md#java

bonus: since already installed, you don't have to "pay" the setup time + disk usage (and we are already close to the 10Go runner limit)

.github/workflows/amd64_linux_cmake_java.yml
```diff
@@ -32,8 +32,9 @@ jobs:
           sudo apt update
           sudo apt install -y ninja-build ccache swig
           swig -version
+          echo "JAVA_HOME=$JAVA_HOME_21_X64" >> $GITHUB_ENV
       - name: Check java
-        run: java -version
+        run: $JAVA_HOME/bin/java -version
```
Mizux: you could also have used previously:
```yml
echo "$JAVA_HOME/bin" >> "$GITHUB_PATH"
```
ref: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-system-path

note: thus you should have `java`, `javac` and `jar` in sync...

gchatelet: Fixed in [5e4007f](https://github.com/google/or-tools/pull/4966/commits/5e4007fcddbf5613538a8de7d0ae13df31d7e5dc)

gchatelet:
I ended up removing "Check Java" and only setting `JAVA_HOME` in `GITHUB_ENV`

Mizux (approved):
LGTM

PR #5020, opened 2026-02-03
linear_solver: Fix highs feasible solution extract

kiijeonghoo:
When HiGHS stops with a feasible solution (e.g. time/iteration limit), variable and objective values were not filled, so `getValue()` returned nothing. This change fills them for both OPTIMAL and FEASIBLE.


lperron:
Thanks

Issue #4988, opened 2026-01-15
CpSolver.Solve() callback breaking argument change


ThomSerg:
**What version of OR-Tools and what language are you using?**
Version: main/9.15.6755
Language: Python

**Which solver are you using (e.g. CP-SAT, Routing Solver, GLOP, BOP, Gurobi)**
CP-SAT

With the latest release of the Python API, there seem to have been an unintended breaking change to the `CpSolver.Solve()` arguments. `.Solve()` has been marked deprecated in favor of `.solve()`, with the former still forwarding the calls to the latter. But it seems like a mistake has been made here, accidentally renaming argument `solution_callback` to `callback`.

Minimal reproducible example:
```python
from ortools.sat.python.cp_model import CpModel, CpSolver, CpSolverSolutionCallback

class MyCallback(CpSolverSolutionCallback):
    def on_solution_callback(self):
        print("Solution found!")

# Create model
model = CpModel()
x = model.new_int_var(0, 5, 'x')
y = model.new_int_var(0, 5, 'y')
model.add(x < y)

# Solve with callback
solver = CpSolver()
solver.parameters.enumerate_all_solutions = True
solver.Solve(model, solution_callback=MyCallback()) # <- instead of a deprecation warning, has become a breaking change
solver.solve(model, solution_callback=MyCallback()) # <- still works fine
```
```bash
TypeError: CpSolver.Solve() got an unexpected keyword argument 'solution_callback'
```

Mizux:
Workaround use `callback` as naming arg when using `Solve()`.

```py
    status = solver.Solve(model, callback=MyCallback())
    status = solver.solve(model, solution_callback=MyCallback())
```
or without naming
```py
    status = solver.Solve(model, MyCallback())
    status = solver.solve(model, MyCallback())
```

Mizux:
fixed on main
https://github.com/google/or-tools/blob/ecbd3d2ac8f0a87159b234175fdccf84847693d4/ortools/sat/python/cp_model.py#L2023-L2027

PR #5065, opened 2026-02-25
Fix issue #4988: CpSolver.Solve() callback breaking argument change

philipp-u:
This PR fixes an issue introduced in commit 4a2de332ce318b98d4489bc6265c84e270bcb09f and raised in #4988 where a breaking change was made whilst improving the quality of the CP-SAT python codebase. The change is very minor, and introduces a small test as an additional thought. I appreciate the method is deprecated so the test may feel more like bloat in this case, I am happy to adjust either way.


philipp-u:
I see this is already addressed in `main`

PR #5094, opened 2026-03-18
Update pybind11 from 2.13.6 to 3.0.1

Mizux:
side note: when using pybind11 3.0.2 the following test will fails
- python_linear_solver_model_builder_test (Failed)
- python_sat_cp_model_helper_test (Failed)
- python_sat_cp_model_test (Failed)

to test with cmake:
```sh
# first hack cmake/dependencies/CMakeLists.txt to use pybind v3.0.2 TAG
cmake -S. -Bbuild_python -DBUILD_PYTHON=ON
cmake --build build_python -j 12
(cd build_python && ctest -C Release -R "python_linear_solver_model_builder
(cd build_python && ctest -C Release -R "python_sat_cp_model.*")
```

to investigate...


Mizux:
note: With pybind 3.0.2 you can't use `py::keep_alive<>` inside a `def_property_readonly` and pybind11 3.0.2 has an assert to check that.
ref: https://github.com/pybind/pybind11/commit/af796d0a99f0cbd9aebb10591257c41a56811cf6
ref: https://pybind11.readthedocs.io/en/stable/advanced/functions.html#return-value-policies

PR #5147, opened 2026-04-26
fix(routing): respect use_relocate_neighbors search parameter in GetNeighborhoodOperators

### LINKED ISSUE #5132: use_relocate_neighbors is bypassed in GetNeighborhoodOperators
Version: stable
Language: C++

Method `LocalSearchOperator* Model::GetNeighborhoodOperators()`

```c++
if (!pickup_delivery_pairs_.empty() and
      search_parameters.local_search_operators().use_relocate_neighbors() ==
          BOOL_TRUE) {
    operators.push_back(local_search_operators_[RELOCATE_NEIGHBORS]);
  }
```
**What did you see instead?**
```c++
if (!pickup_delivery_pairs_.empty() ||
      search_parameters.local_search_operators().use_relocate_neighbors() ==
          BOOL_TRUE) {
    operators.push_back(local_search_operators_[RELOCATE_NEIGHBORS]);
  }
```

**Issue**

I cannot controll the neighborhood operator, since one of them (see above) is bypassing the search parameters.

KasimVali2207:
Fixes #5132

## Problem
In `Model::GetNeighborhoodOperators()`, the `RELOCATE_NEIGHBORS` operator was
being unconditionally added whenever `pickup_delivery_pairs_` was non-empty,
completely ignoring the `use_relocate_neighbors` search parameter.

This was caused by incorrect use of `||` (OR) instead of `&&` (AND) in the
condition check around `routing.cc` line ~5255.

## Fix
Changed the logic from `||` to `&&` so that `RELOCATE_NEIGHBORS` is only added
when **both** conditions are true:
- `pickup_delivery_pairs_` is non-empty
- `use_relocate_neighbors` is enabled in the search parameters

## Files Changed
- `ortools/routing/routing.cc` — fixed boolean condition in `GetNeighborhoodOperators()`

### DIFF
--- a/ortools/constraint_solver/routing.cc
+++ b/ortools/constraint_solver/routing.cc
@@ -5170,7 +5170,9 @@ LocalSearchOperator* RoutingModel::GetNeighborhoodOperators(
     CP_ROUTING_PUSH_OPERATOR(EXCHANGE, exchange);
     CP_ROUTING_PUSH_OPERATOR(CROSS, cross);
   }
-  if (!pickup_delivery_pairs_.empty() ||
+  if ((!pickup_delivery_pairs_.empty() &&
+       search_parameters.local_search_operators().use_relocate_neighbors() !=
+           BOOL_FALSE) ||
       search_parameters.local_search_operators().use_relocate_neighbors() ==
           BOOL_TRUE) {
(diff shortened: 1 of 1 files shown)


Mizux:
we have fixed it internally by removing the pickup_pair check, will try to export it today...

PR #5158, opened 2026-05-08
Xpress: empty bound check not working -- fixed

### LINKED ISSUE #5085: Xpress: empty bound check not working
For SCIP and GLPK the following test pass since integer round leading to empty bound is correctly managed

We should try to improve xpress_solver to have the same behaviour...

https://github.com/google/or-tools/blob/4336f9f3aeedd5462ee0008116b1a2a1e7d633ef/ortools/math_opt/solver_tests/mip_tests.cc#L141-L157

ref: GLPK implem...
https://github.com/google/or-tools/blob/4336f9f3aeedd5462ee0008116b1a2a1e7d633ef/ortools/math_opt/solvers/glpk_solver.cc#L1782-L1809
https://github.com/google/or-tools/blob/4336f9f3aeedd5462ee0008116b1a2a1e7d633ef/ortools/math_opt/solvers/glpk_solver.cc#L1090-L1098

Tahira-2:
Title:   Xpress: handle empty integer bounds (fixes #5085)

Problem
The MathOpt parametrised test FractionalBoundsContainNoInteger skipped Xpress because integer variables with unrounded bounds like [0.5, 0.6] are rejected by Xpress at variable creation: Xpress rounds the bounds on input, sees [1, 0], and refuses to add the column. SCIP, GLPK, and HiGHS instead return TerminationReason::kInfeasible.

Fix
Mirrors GLPK's EmptyIntegerBoundsResult() pattern:

In AddNewVariables, detect integer variables where lb <= ub && ceil(lb) > floor(ub). For those, substitute [0, 0] in the bounds passed to Xpress (so it accepts the column) and remember the original bounds in a new empty_integer_bounds_vars_ member.
In Solve(), after the existing ListInvertedBounds check, short-circuit with ResultForIntegerInfeasible(...) if any such variable was recorded — without invoking the optimizer. The original lb/ub flow into the user-facing detail message.
Remove the GTEST_SKIP for kXpress in the test.
Files touched
ortools/math_opt/solvers/xpress_solver.{h,cc} — detection + early-return
ortools/math_opt/solvers/BUILD.bazel — added //ortools/math_opt/core:empty_bounds dep
ortools/math_opt/solver_tests/mip_tests.cc — re-enable the test for Xpress
CMake's solver target is built via a recursive glob, so no CMake change is needed.

Test plan
 CI: compile + run all non-Xpress solver tests (covered automatically)
 Local Xpress run of FractionalBoundsContainNoInteger — needs an Xpress license; cc @Mizux for verification

### DIFF
--- a/ortools/math_opt/solvers/xpress_solver.h
+++ b/ortools/math_opt/solvers/xpress_solver.h
@@ -18,6 +18,7 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <vector>
 
 #include "absl/base/nullability.h"
 #include "absl/container/linked_hash_map.h"
@@ -227,6 +228,18 @@ class XpressSolver : public SolverInterface {
   bool nonbinary_indicator_ = false;
   bool is_multiobj_ = false;
   bool is_mip_ = false;
+
+  // Tracks integer variables whose unrounded bounds are non-empty (lb <= ub)
+  // but whose rounded integer bounds are empty (ceil(lb) > floor(ub)). Xpress
+  // rejects creation of such variables, so the bounds passed to Xpress are
+  // replaced with [0, 0] and these entries are kept here. Solve() detects a
+  // non-empty list and returns an infeasible result instead of invoking Xpress.
+  struct EmptyIntegerBoundsVar {
+    VarId id;
+    double lb;
+    double ub;
+  };
+  std::vector<EmptyIntegerBoundsVar> empty_integer_bounds_vars_;
   // Results of the last solve
   int primal_sol_avail_ = XPRS_SOLAVAILABLE_NOTFOUND;
   int dual_sol_avail_ = XPRS_SOLAVAILABLE_NOTFOUND;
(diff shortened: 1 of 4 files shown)


ortools/math_opt/solvers/xpress_solver.h
```diff
@@ -227,6 +228,18 @@ class XpressSolver : public SolverInterface {
...
   bool is_multiobj_ = false;
   bool is_mip_ = false;
+
+  // Tracks integer variables whose unrounded bounds are non-empty (lb <= ub)
+  // but whose rounded integer bounds are empty (ceil(lb) > floor(ub)). Xpress
+  // rejects creation of such variables, so the bounds passed to Xpress are
+  // replaced with [0, 0] and these entries are kept here. Solve() detects a
+  // non-empty list and returns an infeasible result instead of invoking Xpress.
+  struct EmptyIntegerBoundsVar {
+    VarId id;
```
soppera, line 238: **Optional:** you may want to use `ABSL_REQUIRE_EXPLICIT_INIT` from [`#include "absl/base/attributes.h"`](https://github.com/abseil/abseil-cpp/blob/a1b9918b18364077b32a820862c63e14c1edb70b/absl/base/attributes.h#L717) on fields that should have a value specified (especially when those fields would not get a value).

soppera (approved):
Thanks!

Mizux:
cherry picked, rebased on `main` and fix conflict  in https://github.com/google/or-tools/pull/5408

Mizux:
merged via #5408

Issue #4924, opened 2025-11-24
MacOS(Python): SWIG 4.4.0 regression


Mizux:
When migrating from SWIG 4.3.1 to 4.4.0 we got some issue on few tests...

```sh
cmake -S. -Bbuild -DBUILD_PYTHON=ON && cmake --build build --config Release -j 4
(cd build && ctest -V -R "python.*issue46")
```

Observed:
```
...
677: TypeError: SWIG director type mismatch in output value of type 'operations_research::Decision *'
```

src: https://github.com/google/or-tools/blob/main/examples/tests/issue46.py

https://github.com/google/or-tools/blob/fcc7f8567f9b6b2616b7547259d1e2368ef66f3d/ortools/constraint_solver/python/constraint_solver.swig#L118
https://github.com/google/or-tools/blob/fcc7f8567f9b6b2616b7547259d1e2368ef66f3d/ortools/constraint_solver/python/constraint_solver.swig#L133
https://github.com/google/or-tools/blob/fcc7f8567f9b6b2616b7547259d1e2368ef66f3d/ortools/constraint_solver/python/constraint_solver.swig#L1273

Could be related to https://github.com/swig/swig/issues/3279

Mizux:
https://github.com/google/or-tools/blob/fcc7f8567f9b6b2616b7547259d1e2368ef66f3d/ortools/util/python/vector.swig#L196-L242

Mizux:
currently mitigated by forcing the use of swig 4.3.1 on macos cmake python jobs

PR #5186, opened 2026-05-21
java: export util/java from google3

Mizux:


Mizux:
I think the cmake x macos x python error are OK since the last homebrew swig package is broken
see: 
* https://github.com/google/or-tools/issues/4924

PR #5223, opened 2026-06-12
Fix formatting in scheduling docs for interval variables

ClaytonPassmore:
### What / Why

Just a very small formatting fix!

A line break was causing part of a single tick code expression to be interpreted as a markdown bullet point. See the preview of the diff.


Mizux:
please send a PR against main branch

PR #5232, opened 2026-06-18
[CBC interface] Support setting CBC hints

comick:
Hi, the PR adds support for setting CBC hints and making use of them.

I could not find a test within ortools that measures hints impact consistently across solvers (that's implementation dependent and requires looking into either logs, comparing solver time or look into internal iterations and would take a considerable effort).
The included (java only) tests help verifying correctness of the solution when hints are provided.
This also covers other solvers whose implementations I did not touch. I could not find other tests covering hint/no-hint correctness comparisons.

I'd be happy to get some feedback and adjust the PR accordingly, please let me know what you think.

### DIFF
--- a/ortools/javatests/com/google/ortools/linearsolver/LinearSolverTest.java
+++ b/ortools/javatests/com/google/ortools/linearsolver/LinearSolverTest.java
@@ -653,4 +653,229 @@ public final class LinearSolverTest {
 
     assertFalse(solver.setNumThreads(4));
   }
+
+  private void runSolveWithHint(MPSolver.OptimizationProblemType problemType) {
+    if (!MPSolver.supportsProblemType(problemType)) {
+      return;
+    }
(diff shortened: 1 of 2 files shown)


comick:
Hi there, do you think it is possible to take a look at this PR?
I'd be happy to get some feedback and adjust the PR accordingly, please let me know what you think.

comick:
Applied clang-format that made the previous workflow run fail.

comick:
After testing the change on couple bigger problems, it turns out that calling both `setBestSoluton` and `setHotstartSolution` did not work as intended performance-wise.
Using only `setBestSolution` with a pre-computed objective value, I could observe a speedup between 1.5x and 1.8x (compared to cold start), which is an improvement worth setting hints for if available. The speedup did not materialize if the objective value was not pre-computed.

comick:
hi there, wondering if there's anything I can do to get this through.

PR #5242, opened 2026-07-02
examples/python: add tree ensemble global optimisation via CP-SAT

adrianhtt:
I've been using OR-Tools to optimise process setpoints where the objective is captured by a trained tree ensemble surrogate, and I couldn't find an example of this pattern in the repo. Adding one here in case it's useful to others.

The technique is from Misic (2020) — a gradient-boosted ensemble is piecewise-constant over a discrete input domain, so its global minimum can be found exactly as an integer program rather than by heuristic search.

**Encoding sketch**

For each tree:
- One-hot binary vars select each tunable feature's candidate value (`AddExactlyOne` per feature).
- A binary var per leaf, `AddExactlyOne` per tree.
- Linking constraint: a leaf can only be active if every split on its root-to-leaf path is consistent with the chosen feature values.

The objective is just the sum of active leaf values (scaled to integers for CP-SAT).

**The example**

Two trees, three integer setpoints, 36 combinations total — small enough that a brute-force verifier can confirm the CP-SAT result is the true global optimum. No external ML library needed; the tree structure is supplied as a Python dict (the same form you'd get from `model.get_booster().trees_to_dataframe()` or a similar export).

**Files**

- `tree_ensemble_optimization_sat.py` — the example
- `tree_ensemble_optimization_sat_py_test.bintest` — test runner
- `BUILD.bazel` and `CMakeBazel.txt` — entries added alphabetically alongside the other `*_sat` examples

Reference: Misic, V.V. (2020). Optimization of Tree Ensembles. *Operations Research*, 68(5), 1605-1624.

### DIFF
--- a/examples/python/BUILD.bazel
+++ b/examples/python/BUILD.bazel
@@ -573,6 +573,23 @@ bintest(
 )
 
 py_binary(
+    name = "tree_ensemble_optimization_sat_py3",
+    srcs = ["tree_ensemble_optimization_sat.py"],
+    main = "tree_ensemble_optimization_sat.py",
+    deps = [
+        requirement("absl-py"),
+        "//ortools/sat/python:cp_model",
+    ],
+)
+
+bintest(
+    name = "tree_ensemble_optimization_sat_py_test",
+    size = "small",
+    srcs = [":tree_ensemble_optimization_sat_py_test.bintest"],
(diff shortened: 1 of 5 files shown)


lperron:
Thanks for the contribution

PR #5243, opened 2026-07-05
python: declare free-threading support in graph modules

joshlk:
Adds `pybind11::mod_gil_not_used()` to the three `ortools.graph.python` modules (`min_cost_flow`, `max_flow`, `linear_sum_assignment`).

OR-Tools has shipped free-threaded (cp313t/cp314t) Linux wheels since 9.12, but because no extension module declares `Py_mod_gil_not_used`, importing e.g. `ortools.graph.python.min_cost_flow` on a free-threaded interpreter emits a `RuntimeWarning` and re-enables the GIL process-wide, which defeats the purpose of the free-threaded build.

These three modules are stateless pybind11 bindings around `SimpleMinCostFlow` / `SimpleMaxFlow` / `SimpleLinearSumAssignment`, which hold all state per instance with no global/static mutable state.

Motivation: [k-means-constrained](https://github.com/joshlk/k-means-constrained) uses `SimpleMinCostFlow` and is adding free-threaded (3.14t) wheels for true multi-threaded solving; this declaration is what currently blocks the GIL staying disabled.

### DIFF
--- a/ortools/graph/python/linear_sum_assignment.cc
+++ b/ortools/graph/python/linear_sum_assignment.cc
@@ -20,7 +20,7 @@ using ::operations_research::SimpleLinearSumAssignment;
 using NodeIndex = ::operations_research::SimpleLinearSumAssignment::NodeIndex;
 using ::pybind11::arg;
 
-PYBIND11_MODULE(linear_sum_assignment, m) {
+PYBIND11_MODULE(linear_sum_assignment, m, pybind11::mod_gil_not_used()) {
   pybind11::class_<SimpleLinearSumAssignment> slsa(m,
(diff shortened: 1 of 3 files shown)


joshlk:
CI note: the two failing checks are unrelated to this change and also fail on `main`:
* clang-format flags `ortools/lp_data/lp_types.cc`
* `ortools/linear_solver/linear_solver.cc` (untouched here; same failure on #5241), and Bazel macOS aborts in `//ortools/routing/samples:vrp_resources_py3` with a protobuf shutdown mutex deadlock (also failed on #5240/#5241).

All jobs exercising this change pass, including the graph Python sample tests.

PR #5248, opened 2026-07-08
fix(python): statically link and hide bundled HiGHS to avoid symbol clash with highspy (#5246)

### LINKED ISSUE #5246: Incompatibility between highspy and or-tools
Version: ortools==9.14.6206 from PyPI
Language: Python

**Which solver are you using (e.g. CP-SAT, Routing Solver, GLOP, BOP, Gurobi)**

None.

**What operating system (Linux, Windows, ...) and version?**

Linux, GitHub CI Ubuntu and ArchLinux

**What did you do?**

1. Install CVXPY with its default solvers (including the `highspy` package)
2. import CVXPY (or just highs)
3. Solve a problem in CVXPY with GLOP or PDLP from OR Tools (or call them directly)

```
  uv run --no-project --python 3.12 \
    --with 'ortools==9.14.6206' \
    --with 'highspy==1.15.1' \
    python -c 'import highspy; from ortools.linear_solver import pywraplp; print(pywraplp.Solver.CreateSolver("GLOP"))'
```

Successful creation of the solver.

**What did you see instead?**

```
ImportError: /home/ptnobel/.cache/uv/archive-v0/Fu73w6zBZdQnl2ZR/lib/python3.12/site-packages/ortools/linear_solver/../../ortools/.libs/libortools.so.9: undefined symbol: _Z19setLocalOptionValueRK15HighsLogOptionsRKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEERS_RSt6vectorIP12OptionRecordSaISD_EES7_
```

Make sure you include information that can help us debug (full error message, model Proto).

**Anything else we should know about your project / environment**

This issue affects all `cvxpy` + or-tools users since the latest HIGHS release.

My best attempt at a diagnosis is that OR Tools's bundled HIGHS has a symbol conflict with HIGHS from HIGHSpy. One fix I want to do on the CVXPY side is stop eagerly loading libraries, but we would like to allow users to switch solvers easily etc. and so we would appreciate if our dependencies can be simultaneously imported (which we do in CI for example, and we have explicit methods designed to fall through to other solvers when one fails that rely on the ability to simultaneously import, etc.).

--- PTN, CVXPY Maintainer

CC: @jahall @galabovaa

aryan5v:
Fixes #5246.

## Problem

Any Python process that imports both `ortools` and `highspy` (>= 1.15.1) crashes on solver creation:

```
ImportError: .../ortools/.libs/libortools.so.9: undefined symbol:
  _Z19setLocalOptionValueRK15HighsLogOptions...   # setLocalOptionValue(HighsLogOptions const&, ...)
```

Reproducer (released wheel):

```bash
uv run --no-project --python 3.12 \
  --with 'ortools==9.14.6206' --with 'highspy==1.15.1' \
  python -c 'import highspy; from ortools.linear_solver import pywraplp; print(pywraplp.Solver.CreateSolver("GLOP"))'
```

This breaks **all** `cvxpy` + `or-tools` users, since cvxpy eagerly co-imports solver backends. It is not HiGHS-specific at the call site — the whole `libortools.so` fails to load, so GLOP/PDLP/etc. all fail too.

## Root cause

On Linux the wheel ships HiGHS as a **separate shared library** (`libhighs.so`, in `ortools/.libs/`), and `libortools.so` carries **undefined** HiGHS references resolved at load time. OR-Tools bundles HiGHS v1.12.0; `highspy` bundles v1.15.1, which renamed/removed `setLocalOptionValue`. Both copies export HiGHS symbols with **default ELF visibility** into the global dynamic scope. When `highspy` is imported first, its symbols occupy global scope; loading `libortools.so` afterward interposes its HiGHS references onto highspy's incompatible copy, where the symbol no longer exists → `undefined symbol` at `dlopen` time. `auditwheel repair` renames bundled `.so` files but does not mangle symbols, so it does not prevent the collision.

macOS (two-level namespace) and Windows (explicit `OR_EXPORT`) are unaffected, so every change is guarded `if(UNIX AND NOT APPLE)`.

## Fix

Statically link the bundled HiGHS into `libortools.so` and localize its symbols, Linux-only:

- **`cmake/dependencies/CMakeLists.txt`** — build the FetchContent HiGHS with `BUILD_SHARED_LIBS=OFF` (saved/restored around the fetch) so it produces `libhighs.a`, folded into `libortools.so`. HiGHS references become internally resolved — no longer undefined, so there is nothing for another HiGHS copy to interpose onto.
- **`cmake/cpp.cmake`** — add `target_link_options(ortools PRIVATE "LINKER:--exclude-libs,libhighs.a")` so the statically-linked HiGHS symbols become `STB_LOCAL` in `libortools.so`'s dynamic symbol table (not exported, not interposable). Targeted at the HiGHS archive specifically to avoid over-hiding abseil/protobuf/OR-Tools public symbols.
- **`cmake/python.cmake`, `cmake/java.cmake`, `cmake/dotnet.cmake`** — a static library has no SONAME, so the `libhighs.so` copy-into-package step is skipped when HiGHS is static (via an `is_highs_shared` type check); the copy command collapses to a no-op.

This is immune to import order and to whatever HiGHS version another package bundles.

## Scope

- HiGHS only. Other bundled solvers (SCIP, CoinOR, GLPK) are not touched — no reported conflict, and CoinOR in particular is a cluster of interdependent shared libs where static-linking would be riskier.
- CMake only. The PyPI wheels are built via CMake; the Bazel build links HiGHS differently and is not used for the wheels, so it is out of scope here.

## Validation (Linux, x86_64, Python 3.12)

Built the unmodified branch and ran the reproducer against the locally-built wheel:

- Released-wheel bug reproduced in both import orders (baseline).
- Build succeeds with no patches; HiGHS built static (`build/lib/libhighs.a`, no `libhighs.so*`).
- `nm -D` shows no exported `setLocalOptionValue`; `readelf -d` shows no `NEEDED` entry for HiGHS; `setLocalOptionValue` is present only as a local `t` symbol.
- Wheel `.libs/` contains no `libhighs`.
- Both import orders with `highspy==1.15.1` now pass; GLOP, PDLP, and HIGHS all solve.

CI note: `.github/workflows/amd64_linux_cmake_python.yml` exercises the changed Linux wheel build path.


cmake/java.cmake
```diff
@@ -477,8 +478,8 @@ add_custom_command(
     ${JAVA_RESSOURCES_PATH}/${JAVA_NATIVE_PROJECT}/
 
   COMMAND ${CMAKE_COMMAND} -E
-    $<IF:$<BOOL:${BUILD_HIGHS}>,copy,true>
-    $<${need_unix_highs_lib}:$<TARGET_SONAME_FILE:highs::highs>>
+    $<IF:$<OR:${need_unix_highs_lib},${need_windows_highs_lib}>,copy,true>
+    $<${need_unix_highs_lib}:$<${is_highs_shared}:$<TARGET_SONAME_FILE:highs::highs>>>
```
Mizux: `need_unix_highs_lib` already embed if `is_high_shared` => i don't think you need it here.

cmake/python.cmake
```diff
@@ -638,8 +639,8 @@ add_custom_command(
     ${PYTHON_PROJECT}/.libs
 
   COMMAND ${CMAKE_COMMAND} -E
-    $<IF:$<BOOL:${BUILD_HIGHS}>,copy,true>
-    $<${need_unix_highs_lib}:$<TARGET_SONAME_FILE:highs::highs>>
+    $<IF:$<OR:${need_unix_highs_lib},${need_windows_highs_lib}>,copy,true>
+    $<${need_unix_highs_lib}:$<${is_highs_shared}:$<TARGET_SONAME_FILE:highs::highs>>>
```
Mizux: same here

aryan5v: Removed the redundant is_highs_shared wrapper

PTNobel:
FYI, we're seeing what I think is the same issue on MacOS too

https://github.com/cvxpy/cvxpy/actions/runs/28965926925/job/85948935248?pr=3447#step:11:1037

PR #5256, opened 2026-07-15
Do not redefine absl::SourceLocation when Abseil provides it

daeho-ro:
Abseil LTS 20260526.0 introduced absl/types/source_location.h, which defines absl::SourceLocation. ortools/base/source_location.h also defines absl::SourceLocation unconditionally, causing 'reference to SourceLocation is ambiguous' build errors (e.g. in ortools/math_opt/cpp/message_callback.h) when building against that Abseil version.

Guard the vendored polyfill with __has_include and use Abseil's absl::SourceLocation when the header is present, mapping ABSL_LOC to absl::SourceLocation::current() in that case.

### DIFF
--- a/ortools/base/source_location.h
+++ b/ortools/base/source_location.h
@@ -32,6 +32,11 @@
 
 #include "absl/base/config.h"
 
+// Abseil >= LTS 20260526 provides absl::SourceLocation; use it when available.
+#if defined(__has_include) && __has_include("absl/types/source_location.h")
+#include "absl/types/source_location.h"
+#else
+
 namespace absl {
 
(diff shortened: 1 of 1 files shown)


Mizux:
1) `stable` branch (targeting last release aka v9.15) **do not** target abseil-cpp 20260526.0
https://github.com/google/or-tools/blob/98c165af62df62b3056c2ee0fca66b24e79097cb/MODULE.bazel#L41

2) our dev branch is `main` so please only provide PR for this branch...

3) abseil-cpp 20260526.0 bump has been already done on `main`.

PR #5260, opened 2026-07-18
Validate indices in set cover file readers to prevent OOB access

TristanInSec:
## Summary

The ORLIB SCP/Rail readers and solution text/proto readers accept file-derived integer values as array indices without validating that they are non-negative and in range.

- `ReadOrlibScp`: 1-indexed subset values are converted to 0-indexed via `ParseNextInteger() - 1`. A file value of `0` produces `SubsetIndex(-1)`.
- `ReadOrlibRail`: Same pattern for element indices.
- `ReadSetCoverSolutionText`: 0-indexed subset values are used directly without range checking.
- `ReadSetCoverSolutionProto`: Same pattern.

When the signed `BaseInt` (`int32_t`) is negative, `StrongVector::operator[]` casts it to `size_t`, producing a massive offset that causes out-of-bounds heap access. The existing guard in `AddElementToSubset` (`subset >= num_subsets()`) does not catch negative values because `-1 >= 0` evaluates to `false` for signed comparison.

## Changes

- Add `CHECK_GE` validation in each reader function to reject invalid indices from file input before they reach the model.
- Add `CHECK_LT` for solution readers to validate indices are within the declared range.
- Add defense-in-depth `CHECK_GE` in `AddElementToSubset` to catch negative element/subset indices from any caller.

## Affected files

- `ortools/set_cover/set_cover_reader.cc` (4 reader functions)
- `ortools/set_cover/set_cover_model.cc` (`AddElementToSubset`)

### DIFF
--- a/ortools/set_cover/set_cover_model.cc
+++ b/ortools/set_cover/set_cover_model.cc
@@ -290,6 +290,8 @@ void SetCoverModel::SetSubsetCost(SubsetIndex subset, Cost cost) {
 }
 
 void SetCoverModel::AddElementToSubset(BaseInt element, BaseInt subset) {
+  CHECK_GE(element, 0) << "Invalid element index: " << element;
+  CHECK_GE(subset, 0) << "Invalid subset index: " << subset;
   elements_in_subsets_are_sorted_ = false;
   if (subset >= num_subsets()) {
     num_subsets_ = subset + 1;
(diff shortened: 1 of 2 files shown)


Mizux:
please send PR against `main` branch our devlopment branch, `stable` is just pointing to last release tag

TristanInSec:
> please send PR against `main` branch our devlopment branch, `stable` is just pointing to last release tag

Done. Thanks.

bdb-git (approved):

Mizux:
Please validate the Google CLA so we can merge your changes

TristanInSec:
> Please validate the Google CLA so we can merge your changes

It should pass now, thanks!

PR #5267, opened 2026-07-22
Add more util tests

gchatelet:


ortools/util/BUILD.bazel
```diff
@@ -673,3 +1073,43 @@ cc_library(
         "@rules_cc//cc/runfiles",
     ],
 )
+
+cc_test(
+    name = "random_subset_test",
+    srcs = ["random_subset_test.cc"],
+    copts = ORTOOLS_TEST_COPTS,
+    linkopts = ORTOOLS_DEFAULT_LINKOPTS,
+    tags = ["noci"],
```
Mizux, line 1082: since this one is noci, we should also disable it in CMake...
ref: https://github.com/google/or-tools/blob/main/ortools/util/CMakeLists.txt#L62

possible implem (to add after the loop and name it `cxx_util_random_subset_test`): https://github.com/google/or-tools/blob/62b21b33a97533f376a4ccf3fe663270cfed2a85/ortools/algorithms/CMakeLists.txt#L58-L60

PR #5319, opened 2026-08-30
tools: remove pkg_resources from check_python_deps.py

stephen-huan:
Continuation of https://github.com/google/or-tools/pull/5032. Note that the error is not unique to alpine; it affects any environment with setuptools newer than v82.0.0 (see <https://setuptools.pypa.io/en/latest/deprecated/pkg_resources.html>).


stephen-huan:
Actually the test could be brought back with just the `pkg_resources` import check removed if desired; I've tested it on or-tools v9.15 at least.

PR #5327, opened 2026-09-01
Skip -1 interval in shaving solver model copy

z-inform:
Encountered a consistent segfault that occurred after variables shaving solver performed a model copy and copied unitialized intervals with `-1` values in `interval_mapping`. These values were then used to index an `std::vector` and caused its corruption. 

The PR adds a check for such -1 values in `interval_mapping` and skips them (this approach mirrors similar checks in `ModelCopy::CopyAndMapNoOverlap2D`).


vitor1001:
Makes sense. It took a while for me to understand when it would trigger, but if an optional interval has a enforcement that is now fixed to false, it should happen.

Thanks for the fix!

PR #5345, opened 2026-09-07
Make solver randomization platform independent

PharahAmari:
## Problem

`std::mt19937_64` produces a specified engine sequence, but the mapping performed by `std::uniform_int_distribution`, `std::uniform_real_distribution`, and `std::shuffle` is implementation-defined. As a result, a fixed seed can select different GLOP pivots and CP-SAT branches on libstdc++ and MSVC STL, leading to different search trajectories and solutions.

## Change

- Add explicit platform-independent helpers for bounded integers, doubles in `[0, 1)`, and shuffling.
- Preserve the existing libstdc++ sequence, including its paired-swap shuffle optimization.
- Replace standard-library distributions and shuffles used by GLOP and CP-SAT search randomization.
- Add golden-sequence tests so Linux and native Windows must consume the same random values.

This intentionally preserves current Linux behavior while making native Windows follow the same sequence.

## Validation

- `cxx_util_random_engine_test`
- `cxx_glop_pricing_test`
- `cxx_sat_cp_model_solver_test`
- Native MSVC golden-sequence test
- Native Windows CP-SAT Python wheel smoke test
- Downstream deterministic scheduling regression repeated twice on native Windows and matched the WSL result exactly

### DIFF
--- a/ortools/glop/BUILD.bazel
+++ b/ortools/glop/BUILD.bazel
@@ -63,6 +63,7 @@ cc_library(
     deps = [
         "//ortools/lp_data:lp_types",
         "//ortools/util:bitset",
+        "//ortools/util:random_engine",
         "//ortools/util:stats",
         "@abseil-cpp//absl/log:check",
         "@abseil-cpp//absl/random",
@@ -169,6 +170,7 @@ cc_library(
         ":update_row",
         ":variables_info",
         "//ortools/lp_data:lp_types",
+        "//ortools/util:random_engine",
         "//ortools/util:stats",
         "@abseil-cpp//absl/base:core_headers",
         "@abseil-cpp//absl/log",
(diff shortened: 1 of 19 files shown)


vitor1001:
Hi and thanks for the submittion!

I have mixed feelings about this type of determinism. Once it works, it's very tempting to write unit tests that expect a specific problem to always solve in exactly X seconds, or to always find one particular solution among many. But then, as soon as anything in the CP-SAT codebase changes, for example, fixing a buggy heuristic that was consuming one extra random number, those unit tests will start failing with very puzzling results.

On the other hand, I completely understand how valuable this can be for debugging why something works on one platform but not on another, even if it adds some complexity to our codebase.

The thing that tips the scales for me is that, even with your fixes, we still cannot expect true cross-compiler determinism. Some compilers will use FMA instructions for floating-point math, which calculate intermediate multiplications with higher precision. More generally, it's incredibly hard to guarantee bit-exact results across platforms when code relies on floating-point math, as we do in GLOP. Note that while CP-SAT uses GLOP's solution only as a hint (meaning the correctness of the integer solving isn't compromised), these slight floating-point differences can still change how the solver branches.

PR #5346, opened 2026-09-07
Fix generated proto exports in Windows shared builds

PharahAmari:
## Problem

When generated proto object libraries are compiled as part of a shared MSVC build, `OR_TOOLS_COMPILE_DEFINITIONS` can provide the consumer-side `dllimport` definition while the target also adds its private `dllexport` definition. Both definitions then appear on the compiler command line. If `dllimport` wins by command-line ordering, MSVC rejects definitions of generated static data members with C2491.

## Change

Filter the current generated proto target's consumer-side import macro out of the common compile definitions before applying the target's explicit interface/private import/export definitions.

Non-MSVC and static builds are unchanged.

## Validation

- Configured and built a native Windows shared OR-Tools Python wheel with MSVC.
- Verified generated proto compilation contains only the target's `dllexport` definition.
- Successfully linked `ortools.dll` and all Python extension modules.
- Imported MathOpt protobuf modules and solved a CP-SAT smoke model from the resulting wheel.

### DIFF
--- a/cmake/cpp.cmake
+++ b/cmake/cpp.cmake
@@ -563,7 +563,18 @@ function(generate_proto_library)
     ${PROJECT_BINARY_DIR}
     #$<TARGET_PROPERTY:protobuf::libprotobuf,INTERFACE_INCLUDE_DIRECTORIES>
   )
-  target_compile_definitions(${PROTO_NAME}_proto PUBLIC ${OR_TOOLS_COMPILE_DEFINITIONS})
+  # OR_TOOLS_COMPILE_DEFINITIONS contains the import definition used by
+  # consumers of the generated proto library. Do not also pass it when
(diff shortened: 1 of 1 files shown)


cmake/cpp.cmake
Mizux, line 55: why not removing theses lines (53 to 55) from cpp.cmake instead of filter out later ?

https://github.com/google/or-tools/blob/5d0f4ac0270fe37e12b767a5741f33e8a2c6cb8c/cmake/cpp.cmake#L53-L55
knowing we already have:
https://github.com/google/or-tools/blob/5d0f4ac0270fe37e12b767a5741f33e8a2c6cb8c/cmake/cpp.cmake#L567-L572
and
https://github.com/google/or-tools/blob/5d0f4ac0270fe37e12b767a5741f33e8a2c6cb8c/cmake/cpp.cmake#L602-L604

PR #5376, opened 2026-09-10
fix(linear_solver): map MPSolverResponseStatus to ResultStatus safely (fixes #5372)

### LINKED ISSUE #5372: Java: MPSolver.solve() throws IllegalArgumentException for status 99 when HiGHS stops at a limit

Version: v9.15.6755 (latest release on Maven Central); the analysis below also refers to `main`
Language: Java (JDK 21)

**Which solver are you using (e.g. CP-SAT, Routing Solver, GLOP, BOP, Gurobi)**

HIGHS (HiGHS 1.12.0 as bundled in 9.15.6755), through both `MPSolver` (`ortools/linear_solver`) and `ModelBuilder`/`ModelSolver`.

**What operating system (Linux, Windows, ...) and version?**

Ubuntu Linux 24.04, kernel 6.17, x86-64, OpenJDK 21.0.12

**What did you do?**

1. Unpack the attached directory and run `./run.sh` (needs a JDK and maven; maven is only used to fetch `ortools-java` and its native jar). It compiles and runs the single file `HighsTimeLimitStatus.java`.
2. The reproducer builds a market split instance (Cornuéjols/Dawande): 4 equality rows, 26 binary variables plus 4 continuous slacks, minimizing the sum of the slacks. HiGHS needs about 10 seconds to prove optimality, so a 3 second limit is always hit while an incumbent is already known.
3. The same model is solved three times: with `MPSolver` and a 3 s time limit, with `ModelSolver` and a 3 s time limit, and with `ModelSolver` without a time limit.

HiGHS reports everything that is needed for a proper result. Its own log inside the OR-Tools run says:

```
  Status            Time limit reached
  Primal bound      6
  Dual bound        0
  Solution status   feasible
                    6 (objective)
```

So the expectation is:

* `MPSolver.solve()` returns `FEASIBLE` (or at the very least a valid `ResultStatus` value) and the incumbent is available through `MPVariable.solutionValue()`,
* `ModelSolver.solve()` returns `FEASIBLE` with `hasSolution() == true`.

**What did you see instead?**

1. `MPSolver.solve()` throws:

```
java.lang.IllegalArgumentException: No enum class com.google.ortools.linearsolver.MPSolver$ResultStatus with value 99
    at com.google.ortools.linearsolver.MPSolver$ResultStatus.swigToEnum(MPSolver.java:760)
    at com.google.ortools.linearsolver.MPSolver.solve(MPSolver.java:326)
```

2. `ModelSolver.solve()` returns `UNKNOWN_STATUS` with `hasSolution() == false`, i.e. the incumbent HiGHS found is silently dropped.

3. Without a time limit the very same model solves fine: `OPTIMAL`, objective 6.

Full output of the three runs: `output-ortools-9.15.6755.txt`.

**Analysis**

There problem consists of two parts:
1. the status/solution transfer from Highs to OrTools in C++, already fixed on main but no released yet,
2. an unchecked status cast instead of a specified status in the transfer from Ortools/C++ to Ortools/Java, not yet fixed on main.

Releasing part 1 would solve the specific problem, and fixing part 2 would add more stability in handling unknown status values. 

Line numbers refer to the tags `v9.15` of or-tools and `v1.12.0`/`v1.15.1` of HiGHS; `main` line numbers are from 2026-09-09.

##### Part 1 - status mapping (already fixed on `main`, but in no release yet)

HiGHS returns `HighsStatus::kWarning` from `run()` whenever it stops at a limit, see `Highs::returnFromOptimizeModel` in `highs/lp_data/Highs.cpp` (lines 4672-4680 in HiGHS 1.12.0, lines 4863-4871 in 1.15.1 - the block is identical in both versions):

```cpp
4672       // Finally consider the warning returns
4673     case HighsModelStatus::kTimeLimit:
4674     case HighsModelStatus::kIterationLimit:
4675     case HighsModelStatus::kSolutionLimit:
4676     case HighsModelStatus::kInterrupt:
4677     case HighsModelStatus::kHighsInterrupt:
4678     case HighsModelStatus::kUnknown:
4679       assert(return_status == HighsStatus::kWarning);
4680       break;
```

In `ortools/linear_solver/proto_solver/highs_proto_solver.cc` of v9.15 that warning short-circuits the mapping, so the model status is never looked at (lines 225-236):

```cpp
225   HighsStatus run_status = highs.run();
226   switch (run_status) {
227     case HighsStatus::kError: {
...
232     case HighsStatus::kWarning: {
233       response.set_status_str("Warning HiGHS run()");
234       break;                        // model status is never inspected
235     }
236     case HighsStatus::kOk: {
```

The response therefore keeps the default of the proto field, which is 99 (`ortools/linear_solver/linear_solver.proto`, lines 605-606):

```proto
605   optional /*required*/ MPSolverResponseStatus status = 1
606       [default = MPSOLVER_UNKNOWN_STATUS];
```

On top of that, v9.15 copies the variable values only for `MPSOLVER_OPTIMAL` (`highs_proto_solver.cc:276`), so even a `MPSOLVER_FEASIBLE` response would arrive without a solution.

This part is [#5016 "Wrong status parsed for HiGHS when time_limit is set"](https://github.com/google/or-tools/issues/5016), fixed on `main` by commit `27ebae018` (2026-01-30): only `kError` returns early now, and the model status switch has explicit `kTimeLimit`/`kIterationLimit`/`kInterrupt`/`kSolutionLimit`/`kMemoryLimit` cases (`highs_proto_solver.cc:258-303` on current `main`). Commit `5ab516b43` (2026-02-03, "HiGHS: populate variable values when status is FEASIBLE") added the missing solution transfer (`highs_proto_solver.cc:324-325`). The `v9.15` tag is from 2026-01-09 and thus predates both, and 9.15.6755 is still the newest artifact on Maven Central, so there is currently no released version that contains the fix. A release containing those two commits would be very welcome.

##### Part 2 - the invalid enum value reaching Java (still present on `main`)

Independently of #5016, `MPSOLVER_UNKNOWN_STATUS` (99) is cast into an enum that does not have that value, in `ortools/linear_solver/linear_solver.cc` (lines 1151-1153, the same in v9.15 and on current `main`):

```cpp
1151 absl::Status MPSolver::LoadSolutionFromProto(const MPSolutionResponse& response,
1152                                              double tolerance) {
1153   interface_->result_status_ = static_cast<ResultStatus>(response.status());
```

`MPSolver::ResultStatus` (`ortools/linear_solver/linear_solver.h`) only has `OPTIMAL(0) … NOT_SOLVED(6)`, and the SWIG generated Java enum consequently throws in `MPSolver.java:760`. After #5016 this path is still reachable: the new code sets `MPSOLVER_UNKNOWN_STATUS` explicitly when a limit is reached *without* a feasible solution (`info.primal_solution_status != kSolutionStatusFeasible`), for instance when the time limit is so short that no incumbent has been found yet. A Java caller then gets an `IllegalArgumentException` out of `solve()` instead of a status it can handle.

Suggested fix: do not cast the response status unchecked. Either map every `MPSolverResponseStatus` without a counterpart to an existing value (`ABNORMAL` or `NOT_SOLVED`), or add the missing values (`MPSOLVER_UNKNOWN_STATUS`, `MPSOLVER_CANCELLED_BY_USER`) to `MPSolver::ResultStatus` so that SWIG can convert them.

**Anything else we should know about your project / environment**

* The problem is not specific to time limits: with the solver parameter `mip_max_improving_sols 1` (HiGHS stops with `kSolutionLimit`) the very same exception occurs.
* It is not a HiGHS problem. The standalone HiGHS CLI 1.15.1 reports the same thing as HiGHS 1.12.0 inside OR-Tools - "Time limit reached", "Solution status feasible", primal bound 6, see `output-highs-cli-1.15.1.txt` (`highs market-split.lp --time_limit 3`).
* Attached directory: [ortools-highs-status-bug.zip](https://github.com/user-attachments/files/32047693/ortools-highs-status-bug.zip)
* Files in this directory: 
   * market-split.lp - small MIP to reproduce the problem
   * HighsTimeLimitStatus.java - small java program to reproduce the bug
   * output-highs-cli-1.15.1.txt - output of standalone highs
   * output-ortools-9.15.6755.txt - output of ortools/highs
   * pom.xml - pom to download the ortools dependencies
   * run.sh - run script

jdymitarai:
### Description

Fixes #5372.

When solvers (such as HiGHS, PDLP, or SAT) terminate due to a limit or non-standard condition without a primal solution, or when an unknown status is returned, the response proto contains statuses such as `MPSOLVER_UNKNOWN_STATUS` (value 99), `MPSOLVER_CANCELLED_BY_USER` (value 98), or model invalid parameter codes (84, 85, etc.).

Previously:
- `MPSolver::LoadSolutionFromProto`: `interface_->result_status_ = static_cast<ResultStatus>(response.status());`
- `HighsInterface::Solve`: `result_status_ = static_cast<MPSolver::ResultStatus>(response->status());`
- `PdlpInterface::Solve`: `result_status_ = static_cast<MPSolver::ResultStatus>(response->status());`

Casting `MPSolverResponseStatus` directly into `MPSolver::ResultStatus` causes out-of-range integer values (e.g. 99) to be assigned to `result_status_`. When queried via SWIG in Java (or .NET), this throws an unhandled exception:
```
java.lang.IllegalArgumentException: No enum class com.google.ortools.linearsolver.MPSolver$ResultStatus with value 99
    at com.google.ortools.linearsolver.MPSolver$ResultStatus.swigToEnum(MPSolver.java:760)
    at com.google.ortools.linearsolver.MPSolver.solve(MPSolver.java:326)
```

### Changes
1. **Public Conversion Utility**: Declared and implemented `MPSolverResponseStatusToResultStatus` in `linear_solver.h` / `linear_solver.cc`.
   - Explicitly maps standard statuses (`OPTIMAL`, `FEASIBLE`, `INFEASIBLE`, `UNBOUNDED`, `ABNORMAL`, `MODEL_INVALID`, `NOT_SOLVED`).
   - Extended validation/options errors (`MPSOLVER_MODEL_INVALID_*`, `MPSOLVER_SOLVER_TYPE_UNAVAILABLE`, `MPSOLVER_INCOMPATIBLE_OPTIONS`) map to `MODEL_INVALID`.
   - Unknown/unhandled statuses (`MPSOLVER_UNKNOWN_STATUS`, `MPSOLVER_CANCELLED_BY_USER`, default) safely map to `MPSolver::ABNORMAL` instead of leaking out-of-range enum values.
2. **Standardized Interfaces**:
   - `MPSolver::LoadSolutionFromProto`, `HighsInterface::Solve`, `PdlpInterface::Solve`, and `SatInterface::Solve` now all route through `MPSolverResponseStatusToResultStatus`.
3. **Unit Tests**:
   - Added `ortools/linear_solver/linear_solver_test.cc` verifying bidirectional conversions, unmapped status handling, and `LoadSolutionFromProto` behavior on unknown status values.
   - Updated `ortools/linear_solver/BUILD.bazel` and `CMakeLists.txt` to register `linear_solver_test`.

### DIFF
--- a/ortools/linear_solver/BUILD.bazel
+++ b/ortools/linear_solver/BUILD.bazel
@@ -836,6 +836,19 @@ cc_test(
     ],
 )
 
+cc_test(
+    name = "linear_solver_test",
+    srcs = ["linear_solver_test.cc"],
+    copts = ORTOOLS_TEST_COPTS,
+    linkopts = ORTOOLS_DEFAULT_LINKOPTS,
+    deps = [
+        ":linear_solver",
+        ":linear_solver_cc_proto",
+        "//ortools/base:gmock_main",
+        "@abseil-cpp//absl/status",
+    ],
+)
+
 cc_library(
     name = "solve_mp_model",
     srcs = ["solve_mp_model.cc"],
--- a/ortools/linear_solver/CMakeLists.txt
+++ b/ortools/linear_solver/CMakeLists.txt
@@ -142,4 +142,9 @@ if(BUILD_CXX_TESTING)
 
     add_test(NAME cxx_unittests_xpress_interface COMMAND test_xprs_interface)
   endif()
+
+  add_executable(linear_solver_test linear_solver_test.cc)
+  target_compile_features(linear_solver_test PRIVATE cxx_std_17)
+  target_link_libraries(linear_solver_test PRIVATE ortools::ortools GTest::gtest_main)
+  add_test(NAME cxx_unittests_linear_solver COMMAND linear_solver_test)
 endif ()
--- a/ortools/linear_solver/highs_interface.cc
+++ b/ortools/linear_solver/highs_interface.cc
@@ -150,7 +150,7 @@ MPSolver::ResultStatus HighsInterface::Solve(const MPSolverParameters& param) {
 
   // The solution must be marked as synchronized even when no solution exists.
   sync_status_ = SOLUTION_SYNCHRONIZED;
-  result_status_ = static_cast<MPSolver::ResultStatus>(response->status());
+  result_status_ = MPSolverResponseStatusToResultStatus(response->status());
 
   if (response->status() == MPSOLVER_FEASIBLE ||
       response->status() == MPSOLVER_OPTIMAL) {
--- a/ortools/linear_solver/linear_solver.cc
+++ b/ortools/linear_solver/linear_solver.cc
@@ -809,7 +809,35 @@ MPSolverResponseStatus MPSolver::LoadModelFromProtoInternal(
   return MPSOLVER_MODEL_IS_VALID;
 }
 
-namespace {
+MPSolver::ResultStatus MPSolverResponseStatusToResultStatus(
+    MPSolverResponseStatus status) {
+  switch (status) {
+    case MPSOLVER_OPTIMAL:
+      return MPSolver::OPTIMAL;
+    case MPSOLVER_FEASIBLE:
+      return MPSolver::FEASIBLE;
+    case MPSOLVER_INFEASIBLE:
+      return MPSolver::INFEASIBLE;
+    case MPSOLVER_UNBOUNDED:
+      return MPSolver::UNBOUNDED;
+    case MPSOLVER_ABNORMAL:
+      return MPSolver::ABNORMAL;
+    case MPSOLVER_MODEL_INVALID:
+    case MPSOLVER_MODEL_INVALID_SOLUTION_HINT:
+    case MPSOLVER_MODEL_INVALID_SOLVER_PARAMETERS:
+    case MPSOLVER_SOLVER_TYPE_UNAVAILABLE:
+    case MPSOLVER_INCOMPATIBLE_OPTIONS:
+      return MPSolver::MODEL_INVALID;
+    case MPSOLVER_NOT_SOLVED:
+    case MPSOLVER_MODEL_IS_VALID:
+      return MPSolver::NOT_SOLVED;
+    case MPSOLVER_CANCELLED_BY_USER:
+    case MPSOLVER_UNKNOWN_STATUS:
+    default:
+      return MPSolver::ABNORMAL;
+  }
+}
+
 MPSolverResponseStatus ResultStatusToMPSolverResponseStatus(
     MPSolver::ResultStatus status) {
   switch (status) {
@@ -830,7 +858,6 @@ MPSolverResponseStatus ResultStatusToMPSolverResponseStatus(
   }
   return MPSOLVER_UNKNOWN_STATUS;
 }
-}  // namespace
 
 void MPSolver::FillSolutionResponseProto(MPSolutionResponse* response) const {
   CHECK(response != nullptr);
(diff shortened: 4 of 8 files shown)


lperron:
I imported this manually, and tweaked to work on our internal build system

Thanks for the contrib

PR #5384, opened 2026-09-11
Add Windows ARM64 Python build

ndabas:
As the title says -- mostly just a copy of the existing AMD64 workflow, but with some minor changes for MSVC ARM64.

I have tested these on my fork and the build passes.

### DIFF
--- a/ortools/graph_base/generic_graph_test.cc
+++ b/ortools/graph_base/generic_graph_test.cc
@@ -121,13 +121,13 @@ TEST(GenericGraphTest, BuilderAndGraphAreMovable) {
 TEST(GenericGraphDeathTest, AddMoreNodesThanReserved) {
   GenericGraph<char>::Builder builder(/*num_nodes=*/2, /*num_edges=*/3);
   builder.AddEdge('a', 'b');
-  if constexpr (DEBUG_MODE) {
-    EXPECT_DEATH(builder.AddEdge('c', 'a'), "const_capacities");
-  } else {
-    builder.AddEdge('c', 'a');
-    EXPECT_THAT(std::move(builder).Build().AllNodes(),
-                ElementsAre('a', 'b', 'c'));
-  }
+#ifndef NDEBUG
+  EXPECT_DEATH(builder.AddEdge('c', 'a'), "const_capacities");
+#else
+  builder.AddEdge('c', 'a');
+  EXPECT_THAT(std::move(builder).Build().AllNodes(),
+              ElementsAre('a', 'b', 'c'));
+#endif
 }
 
 TEST(GenericGraphDeathTest, AddMoreEdgesThanReserved) {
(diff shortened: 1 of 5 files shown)


ortools/graph_base/generic_graph_test.cc
```diff
@@ -121,13 +121,13 @@ TEST(GenericGraphTest, BuilderAndGraphAreMovable) {
 TEST(GenericGraphDeathTest, AddMoreNodesThanReserved) {
   GenericGraph<char>::Builder builder(/*num_nodes=*/2, /*num_edges=*/3);
   builder.AddEdge('a', 'b');
-  if constexpr (DEBUG_MODE) {
```
Mizux, line 124: not sure you need this change or we need to change the implem of DEBUG_MODE instead

patches/scip-v10.0.3.patch
```diff
@@ -107,3 +107,18 @@ index c6ce7283..6b6b1fc8 100644
  install(FILES "${PROJECT_BINARY_DIR}${CMAKE_FILES_DIRECTORY}/scip-config.cmake"
                ${PROJECT_BINARY_DIR}/scip-config-version.cmake
          DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/scip)
+diff --git a/src/scip/intervalarith.c b/src/scip/intervalarith.c
```
Mizux, line 110: out of curiosity did you send this change to scip upstream ?

ortools/graph_base/generic_graph_test.cc
ndabas: I got this on [a run without this change](https://github.com/ndabas/or-tools/actions/runs/34584424159/job/103215223536):

```
C:\a\or-tools\or-tools\ortools\graph_base\generic_graph_test.cc(125,5): error C2094: use of undefined label 'gtest_label_125' [C:\a\or-tools\or-tools\build\ortools\graph_base\graph_base_generic_graph_test.vcxproj]
```

Let me know if there is a better way of dealing with this. I believe this would happen on the MSVC x64 builds as well, we're just seeing this now because I am explicitly using a runner with VS2026, while the Windows x64 runners are currently on VS2022. I could try switching it to the VS2022 toolset though if we don't like this patch.

patches/scip-v10.0.3.patch
ndabas: No not yet -- will do though, I need to get my head around scip a bit more.

ortools/graph_base/generic_graph_test.cc
ndabas: BTW the existing Windows AMD64 Python workflow is failing with the exact same error -- for example, see the full logs from https://github.com/google/or-tools/actions/runs/34607377490/job/103289218544

